PR #1, opened 2021-05-25
Setup Liquibase [POL-30]

snwessel:


src/main/resources/db/changelog/changesets/21052021_initial_schema.yaml
```diff
@@ -0,0 +1,109 @@
+databaseChangeLog:
+  - changeSet:
+      id: "1"
+      author: kfeldman, swessel, zarsky
+      changes:
+        - sql:
+            sql: "CREATE TYPE token_type_enum AS ENUM ('document', 'code')"
```
dvoet: I would call this `document_token` and `access_token` (or just `document` and `access` but plain `access` feels weird)

src/main/resources/db/changelog/changesets/21052021_initial_schema.yaml
```diff
@@ -0,0 +1,109 @@
...
+              - column:
+                  name: user_id
+                  type: text
+                  constraints:
+                    nullable: false
+              - column:
+                  name: provider_id
+                  type: text
+                  constraints:
+                    nullable: false
```
dvoet, line 27: need a unique constraint on these 2 together

dvoet (changes requested):

snwessel:
I just updated the enum type names and added the unique constraint as suggested. When I ran the app, it created the constraint successfully! 
image

dvoet (approved):

PR #5, opened 2021-06-07
Health endpoint [POL-32]

snwessel:
Creating a basic health endpoint which checks for a valid database connection and returns an empty 200 response if everything is OK.

->  Created by @zarsky-broad and I


service/src/main/java/bio/terra/externalcreds/PublicApiController.java
```diff
@@ -0,0 +1,35 @@
...
+  }
+
+  @Override
+  @GetMapping("/status")
+  public ResponseEntity<Void> getStatus() {
+    try {
+      if (Objects.requireNonNull(this.jdbcTemplate.getDataSource()).getConnection().isValid(0)) {
+        return new ResponseEntity<>(HttpStatus.OK);
+      } else {
+        return new ResponseEntity<>(HttpStatus.EXPECTATION_FAILED);
```
dvoet: This should not be a 4xx level error. 4xx errors generally means the client did something wrong and can fix it. I think this should be INTERNAL_SERVER_ERROR. If we really want to distinguish this case from the one below I would suggest 503 but 503 is usually indicative of an apache proxy error so I don't recommend it. The real thing that we should use to distinguish the error condition is logging. Add some logging :)

dvoet (changes requested):
We should have a discussion about application layering but given this is simple I will let it slide for now. Fix the error codes please

snwessel:
@dvoet we added a more detailed 200 response (describing which subsystems we check) and updated the error response type. Before we implement logging, we were thinking that we should maybe talk as a group about how we want to set it up. Let us know what you think!

dvoet (approved):

PR #6, opened 2021-06-10
Make some models

zarsky-broad:


service/src/main/java/bio/terra/externalcreds/models/TokenTypeEnum.java
```diff
@@ -0,0 +1,6 @@
+package bio.terra.externalcreds.models;
+
+public enum TokenTypeEnum {
+  DOCUMENT,
+  ACCESS
```
dvoet: these don't match the database enum `CREATE TYPE token_type_enum AS ENUM ('document_token', 'access_token')`

service/src/main/java/bio/terra/externalcreds/models/GA4GHPassport.java
```diff
@@ -0,0 +1,11 @@
+package bio.terra.externalcreds.models;
+
+import java.sql.Timestamp;
+import lombok.Builder;
+
+@Builder
+public class GA4GHPassport {
```
dvoet, line 7: please add `jwt`

dvoet (changes requested):

dvoet (approved):

dvoet (approved):

PR #11, opened 2021-06-22
implement oauth dance [POL-53 POL-36]

dvoet:


service/spring.gradle
```diff
@@ -8,6 +8,9 @@ dependencies {
 	implementation 'org.springframework.boot:spring-boot-starter-jdbc'
 	implementation 'org.springframework.boot:spring-boot-starter-web'
 	testImplementation 'org.springframework.boot:spring-boot-starter-test'
+	implementation 'org.springframework.security:spring-security-oauth2-core'
+	implementation 'org.springframework.security:spring-security-oauth2-client'
+	implementation 'org.springframework.boot:spring-boot-starter-cache'
```
zarsky-broad, line 13: nit: alphabetize within the block, plz

dvoet: well… it is not alphabetized to start with (test comes before web) and does it really add any value?

snwessel (approved):
LGTM! And thanks again for the walkthrough yesterday!

service/spring.gradle
zarsky-broad: The lines are fully alphabetized, starting with their configuration; web is before test because of implementation vs testImplementation. If you wouldn't mind, I prefer to keep them alphabetized, just dumbly, that way, so that there's some consistent order. There's an intellij plugin called [Lines Sorter](https://plugins.jetbrains.com/plugin/5919-lines-sorter) that I use for alphabetizing groups of lines, if you're feeling lazy.

The value is that the deps are in a predictable order when someone is trying to find one or add a new one.

service/src/main/java/bio/terra/externalcreds/config/ProviderConfig.java
```diff
@@ -21,7 +22,14 @@
     private String clientId;
     private String clientSecret;
     private Duration linkLifespan;
-    private String openidConfiguration;
+    private String issuer;
```
zarsky-broad, line 25: Make sure you change the config in vault, please and thank you

service/src/main/java/bio/terra/externalcreds/controllers/OidcApiController.java
```diff
@@ -20,11 +21,24 @@ public OidcApiController(ProviderService providerService) {
...
+  public ResponseEntity<String> getAuthUrl(
+      String provider, List<String> scopes, String redirectUri, String state) {
+    String authorizationUrl =
+        providerService.getProviderAuthorizationUrl(
+            provider, redirectUri, Set.copyOf(scopes), state);
+
+    if (authorizationUrl == null) {
+      return new ResponseEntity<>(HttpStatus.NOT_FOUND);
+    } else {
+      return new ResponseEntity<>(JSONValue.toJSONString(authorizationUrl), HttpStatus.OK);
```
zarsky-broad: I hate that we have to do this to get a JSON string out. Instead, could we either return an object with a single key, or override the default Spring behavior so it always returns a string as json? I'm just thinking of the future when we have more endpoints that return strings, and I'd love to make something generic that we could move to libs so other services could use it too.

I found these:
- https://stackoverflow.com/a/37906098
- https://www.baeldung.com/spring-httpmessageconverter-rest

service/src/main/resources/application.yml
```diff
@@ -26,3 +26,5 @@ spring:
           max-age: 0
           must-revalidate: true
         use-last-modified: false
+
+logging.level.bio.terra.externalcreds: ${EXTERNALCREDS_LOG_LEVEL:debug}
```
zarsky-broad: Would you mind merging this with the logging on line 1? It would look like this:

```yaml
logging:
  level.bio.terra.externalcreds: ${EXTERNALCREDS_LOG_LEVEL:debug}
  pattern.level: '%X{requestId} %5p'
```

service/src/main/java/bio/terra/externalcreds/services/OAuth2Service.java
```diff
@@ -0,0 +1,138 @@
...
+import org.springframework.security.oauth2.core.endpoint.OAuth2AuthorizationResponse;
+import org.springframework.security.oauth2.core.user.OAuth2User;
+import org.springframework.stereotype.Service;
+
+/**
+ * Service that encapsulates all OAuth2 features need by ECM. General flow: 1)
+ * getAuthorizationRequestUri - visit to authenticate with provider 2) authorizationCodeExchange -
+ * using code resulting from authentication, get access and refresh tokens 3) getUserInfo - using
+ * access token call the user info endpoint - called periodically to get updated user info 4)
+ * authorizeWithRefreshToken - get a new access token using a refresh token when required
```
zarsky-broad: To make intellij format it nicely, here's the version with the proper list, as returned after passing it through the formatter, which insists that list items don't need close tags:
```suggestion
 * Service that encapsulates all OAuth2 features need by ECM. General flow:
 *
 * <ol>
 *   <li>getAuthorizationRequestUri - visit to authenticate with provider
 *   <li>authorizationCodeExchange - using code resulting from authentication, get access and
 *       refresh tokens
 *   <li>getUserInfo - using access token call the user info endpoint - called periodically to get
 *       updated user info
 *   <li>authorizeWithRefreshToken - get a new access token using a refresh token when required
 * </ol>
```

service/src/main/java/bio/terra/externalcreds/services/OAuth2Service.java
```diff
@@ -0,0 +1,138 @@
...
+            .redirectUri(redirectUri)
+            .state(state)
+            .build();
+
+    DefaultAuthorizationCodeTokenResponseClient tokenResponseClient =
+        new DefaultAuthorizationCodeTokenResponseClient();
+    OAuth2AuthorizationCodeGrantRequest codeGrantRequest =
+        new OAuth2AuthorizationCodeGrantRequest(
+            providerClient, new OAuth2AuthorizationExchange(authRequest, authResponse));
+    OAuth2AccessTokenResponse tokenResponse =
```
zarsky-broad: Any reason not to just return this directly?

service/src/main/java/bio/terra/externalcreds/services/OAuth2Service.java
```diff
@@ -0,0 +1,138 @@
...
+   * Given an refresh token, get an access token
+   *
+   * @param providerClient identity provider client, see {@link ProviderClientCache}
+   * @param refreshToken
+   * @return token response containing access and refresh tokens, note that if there is a refresh
+   *     token in this response it should replace the original refresh token which is likely invalid
+   */
+  public OAuth2AccessTokenResponse authorizeWithRefreshToken(
+      ClientRegistration providerClient, OAuth2RefreshToken refreshToken) {
+    // the OAuth2RefreshTokenGrantRequest requires an access token to be specified but a valid one
```
zarsky-broad: I'm confused, a valid one what?

service/src/test/java/bio/terra/externalcreds/controllers/OidcApiControllerTest.java
```diff
@@ -0,0 +1,69 @@
...
+    MultiValueMap<String, String> queryParams = new LinkedMultiValueMap<>();
+    queryParams.add("redirectUri", redirectUri);
+    queryParams.addAll("scopes", List.copyOf(scopes));
+    mvc.perform(get("/api/oidc/v1/{provider}/authorization-url", provider).queryParams(queryParams))
+        .andExpect(content().json(JSONValue.toJSONString(result)));
+  }
+
+  @Test
+  void testGetAuthUrl404() throws Exception {
+    String result = "https://test/authorization/uri";
```
zarsky-broad: This line is unused in this test:
```suggestion
```

service/src/test/java/bio/terra/externalcreds/services/OAuth2ServiceTest.java
```diff
@@ -0,0 +1,76 @@
...
+import org.springframework.boot.autoconfigure.SpringBootApplication;
+import org.springframework.boot.builder.SpringApplicationBuilder;
+import org.springframework.security.oauth2.client.registration.ClientRegistration;
+import org.springframework.security.oauth2.core.endpoint.OAuth2AccessTokenResponse;
+import org.springframework.security.oauth2.core.user.OAuth2User;
+import org.springframework.stereotype.Component;
+
+@SpringBootApplication(scanBasePackages = {"bio.terra.externalcreds", "bio.terra.common.logging"})
+@Component
+public class OAuth2ServiceTest {
```
zarsky-broad, line 17: I'd love to come up with a different way to do this, because as it is now someone has to manually run it separately; at the very least I'd like to have it somewhere different in the folder structure. I'll see if I can come up with a proposal.

zarsky-broad (commented):
Some thoughts.

service/spring.gradle
dvoet: I am going to push back harder and say alphabetically ordering this is a waste of time. If `implementation` comes before `testImplementation` then that destroys the value of trying to find anything. People are just going to ctrl-f as these strings are impossible to read anyway.

I don't think there is value in having predictable order here.

service/src/main/java/bio/terra/externalcreds/config/ProviderConfig.java
dvoet: will do after this merges

service/src/main/java/bio/terra/externalcreds/services/OAuth2Service.java
dvoet: meant to say but _not_ a valid one

service/src/test/java/bio/terra/externalcreds/services/OAuth2ServiceTest.java
dvoet: I do not expect to have to run this often. This is just to have something so that when people fiddle with this code they have something easy to to use for quick iterations.

service/src/main/java/bio/terra/externalcreds/controllers/WebMvcConfiguration.java
```diff
@@ -0,0 +1,22 @@
+package bio.terra.externalcreds.controllers;
```
zarsky-broad, line 1: Should this be in `bio.terra.externalcreds.config`?

service/src/test/java/bio/terra/externalcreds/services/OAuth2ServiceTest.java
```diff
@@ -0,0 +1,76 @@
...
+import java.util.Set;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.autoconfigure.SpringBootApplication;
+import org.springframework.boot.builder.SpringApplicationBuilder;
+import org.springframework.security.oauth2.client.registration.ClientRegistration;
+import org.springframework.security.oauth2.core.endpoint.OAuth2AccessTokenResponse;
+import org.springframework.security.oauth2.core.user.OAuth2User;
+import org.springframework.stereotype.Component;
+
+@SpringBootApplication(scanBasePackages = {"bio.terra.externalcreds", "bio.terra.common.logging"})
```
zarsky-broad: We don't need the google formatted logs
```suggestion
@SpringBootApplication(scanBasePackages = "bio.terra.externalcreds")
```

service/src/test/java/bio/terra/externalcreds/services/OAuth2ServiceTest.java
```diff
@@ -0,0 +1,76 @@
+package bio.terra.externalcreds.services;
+
+import bio.terra.common.logging.LoggingInitializer;
```
zarsky-broad:
```suggestion
```

service/src/test/java/bio/terra/externalcreds/services/OAuth2ServiceTest.java
```diff
@@ -0,0 +1,76 @@
+package bio.terra.externalcreds.services;
```
zarsky-broad, line 1: Could we move this to, like, `bio.terra.manualtests`? Just so that everything in `bio.terra.externalcreds` in the test folder is all jUnit tests, which is what I'd expect.

service/src/test/java/bio/terra/externalcreds/services/OAuth2ServiceTest.java
```diff
@@ -0,0 +1,76 @@
...
+@Component
+public class OAuth2ServiceTest {
+
+  @Autowired private ProviderConfig providerConfig;
+  @Autowired private OAuth2Service oAuth2Service;
+  @Autowired private ProviderClientCache providerClientCache;
+
+  public static void main(String[] args) {
+    new SpringApplicationBuilder(OAuth2ServiceTest.class)
+        .initializers(new LoggingInitializer())
```
zarsky-broad:
```suggestion
```

service/src/test/java/bio/terra/externalcreds/controllers/OidcApiControllerTest.java
```diff
@@ -0,0 +1,67 @@
...
+        .andExpect(content().json("[\"fake-provider1\",\"fake-provider2\"]"));
+  }
+
+  @Test
+  void testGetAuthUrl() throws Exception {
+    String result = "https://test/authorization/uri";
+    String provider = "fake";
+    String redirectUri = "fakeuri";
+    Set<String> scopes = Set.of("openid", "email");
+    String state = null;
```
zarsky-broad, line 39: Could we make these `public final` class-level things so we don't have to redefine them for every test? If the queryParams are going to be the same for each test, we could also make a shared `MultiValueMap` and initialize it in `@beforeEach`

zarsky-broad (approved):
Looks good, a few last suggested tweaks.

service/src/main/java/bio/terra/externalcreds/services/OAuth2Service.java
```diff
@@ -0,0 +1,141 @@
...
+   * Given an refresh token, get an access token
+   *
+   * @param providerClient identity provider client, see {@link ProviderClientCache}
+   * @param refreshToken
+   * @return token response containing access and refresh tokens, note that if there is a refresh
+   *     token in this response it should replace the original refresh token which is likely invalid
+   */
+  public OAuth2AccessTokenResponse authorizeWithRefreshToken(
+      ClientRegistration providerClient, OAuth2RefreshToken refreshToken) {
+    // the OAuth2RefreshTokenGrantRequest requires an access token to be specified but not a valid one
```
zarsky-broad: the linter breaks up this line, could you just run it and then commit so we don't wind up with churn?

service/src/test/java/bio/terra/externalcreds/controllers/OidcApiControllerTest.java
dvoet: I understand the desire to do this however this leads to coupled tests that really have nothing to do with each other. Then down the road we have a chunk of test data that feeds a bunch of different tests making it very difficult to change individual tests or the data.

zarsky-broad: makes sense!

service/src/main/java/bio/terra/externalcreds/controllers/WebMvcConfiguration.java
dvoet: I don't think so. I think of the config package as configuration we expose to control ECM. This class is setting up specific controller logic and does not expose any configuration.

service/src/test/java/bio/terra/externalcreds/services/OAuth2ServiceTest.java
dvoet: I like the package structure of the test to mirror the package structure of the thing being tested. That's where I'd go looking for it.

PR #14, opened 2021-06-28
Get provider link [POL-34]

snwessel:
Creating an endpoint to get a provider link. This includes:
- Getting the user id from SAM
- Setting up error handling 
- Figuring out a process for reading from and writing to the DB (and testing that process)


service/src/main/java/bio/terra/externalcreds/services/SamService.java
```diff
@@ -0,0 +1,43 @@
...
+
+  public String getUserIdFromSam() {
+    try {
+      String header = request.getHeader("authorization");
+      if (header == null) throw new UnauthorizedException("User is not authorized");
+      String accessToken = BearerTokenParser.parse(header);
+
+      return samUsersApi(accessToken).getUserStatusInfo().getUserSubjectId();
+    } catch (ApiException e) {
+      throw new ExternalCredsException(e, HttpStatus.INTERNAL_SERVER_ERROR);
```
dvoet: we should handle a 404 response from sam differently. If sam returns a 404, I would be inclined to return a 403 from ECM but I am happy to hear other opinions.

service/src/main/java/bio/terra/externalcreds/services/SamService.java
```diff
@@ -0,0 +1,43 @@
...
+public class SamService {
+
+  private final HttpServletRequest request;
+
+  @Value("${externalcreds.sam-address}")
+  private String samBasePath;
+
+  public SamService(HttpServletRequest request) {
+    this.request = request;
+  }
```
dvoet: can we confirm that a new SamService instance is created for each request?

service/src/test/java/bio/terra/externalcreds/service/LinkedAccountServiceTest.java
```diff
@@ -0,0 +1,39 @@
...
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+import org.mockito.Mockito;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.test.mock.mockito.MockBean;
+
+public class LinkedAccountServiceTest extends BaseTest {
+  @Autowired private LinkedAccountService linkedAccountService;
+
+  @MockBean private LinkedAccountDAO linkedAccountDAO;
```
dvoet: I don't think we should use a mock here, use the real thing. In this particular test it does not really matter but future tests may rely on complex database state making for a complicated and brittle mock.

service/src/test/java/bio/terra/externalcreds/controllers/OidcApiControllerTest.java
```diff
@@ -0,0 +1,55 @@
...
+
+public class OidcApiControllerTest extends BaseTest {
+
+  @Autowired private MockMvc mvc;
+
+  @MockBean private LinkedAccountService linkedAccountService;
+  @MockBean private SamService samService;
+
+  @Test
+  void testGetLink() throws Exception {
```
dvoet: also test 404 response and when sam returns 404

dvoet (changes requested):

dvoet (approved):

service/src/main/java/bio/terra/externalcreds/controllers/OidcApiController.java
```diff
@@ -27,23 +27,23 @@
...
-    this.samService = samService;
+      LinkedAccountService linkedAccountService,
+      ObjectMapper mapper,
+      ProviderService providerService,
+      SamService samService) {
     this.request = request;
+    this.linkedAccountService = linkedAccountService;
     this.mapper = mapper;
+    this.providerService = providerService;
+    this.samService = samService;
```
dvoet: why this change, specifically the reordering?

snwessel: Honestly, I think it may have been accidental. We had been walking through the PR as a group and @zarsky-broad  suggested making these alphabetical, but then we decided it wasn't worth it and went to change it back (but must not have undone all of our changes, hence the new order).

I'd be happy to change it back or make it alphabetical, but I'm not sure if they were in any particular order to begin with.

dvoet: no need to revert, I just want to make sure there is no expectation that our parameter lists are alphabetical

PR #25, opened 2021-07-08
lombokify

zarsky-broad:


service/src/main/java/bio/terra/externalcreds/controllers/SwaggerController.java
```diff
@@ -19,10 +20,10 @@
 
   public SwaggerController() {
     try {
-      InputStream swaggerClientIdStream =
+      @Cleanup
```
zarsky-broad: I _think_ I need this, but it's hard to tell...

snwessel (approved):

service/src/main/java/bio/terra/externalcreds/controllers/SwaggerController.java
dvoet: I think this should just be
```
      try (val reader = new BufferedReader(new InputStreamReader(new ClassPathResource("rendered/swagger-client-id").getInputStream()))) {
        swaggerClientId = reader.readLine();
      }
```

The try syntax will automatically close the reader which closes all nested readers/streams. I think you should use just `readLine()` instead of reading the whole file into a single string… if there is more than 1 line something is probably wrong.

dvoet (approved):
one change then 👍

PR #28, opened 2021-07-21
POL-37 POST provider oauth code

dvoet:


service/src/main/java/bio/terra/externalcreds/util/CompositeBackOffPolicy.java
```diff
@@ -0,0 +1 @@
+
```
snwessel: Is this empty file supposed to be here?

service/src/main/java/bio/terra/externalcreds/controllers/OidcApiController.java
```diff
@@ -62,37 +62,47 @@ private String getUserIdFromSam() {
...
+    var userId = getUserIdFromSam();
+    var linkedAccount = linkedAccountService.getLinkedAccount(userId, provider);
     if (linkedAccount == null) {
       return ResponseEntity.notFound().build();
     }
-    val expTime =
+    var expTime =
         OffsetDateTime.ofInstant(linkedAccount.getExpires().toInstant(), ZoneId.of("UTC"));
-    val linkInfo =
+    var linkInfo =
```
snwessel: I don't think this variable is being used anymore (now that we're calling `getLinkInfoFromLinkedAccount` below).

snwessel (commented):
This looks great overall, just noticed a couple of small things as I was looking through.

zarsky-broad (approved):
LGTMob

snwessel (approved):

PR #29, opened 2021-07-22
proper jku support and more tests

dvoet:


service/src/test/java/bio/terra/externalcreds/services/AuthorizationCodeExchangeTest.java
```diff
@@ -259,18 +304,25 @@ private GA4GHPassport createTestPassport(List<GA4GHVisa> visas) throws JOSEExcep
     return GA4GHPassport.builder().jwt(jwtString).expires(passportExpiresTime).build();
   }
 
-  private String createVisaJwtString(TokenTypeEnum tokenType)
-      throws URISyntaxException, JOSEException {
+  private String createVisaJwtString(GA4GHVisa visa) throws URISyntaxException, JOSEException {
```
dvoet, line 307: Note that I changed this back to taking a visa as a parameter. The intent is to have the jwt match the contents of the visa. While the code in either case currently has the same result the old way does not follow the intent and is ultimately less useful.

Shakespeared (approved):
LGTM

PR #31, opened 2021-07-22
right visa claim structure

dvoet:


snwessel (approved):
LGTM!

service/src/main/java/bio/terra/externalcreds/services/ProviderService.java
```diff
@@ -210,8 +212,13 @@ private GA4GHPassport buildPassport(Jwt passportJwt) {
   }
 
   private GA4GHVisa buildVisa(Jwt visaJwt) {
+    JSONObject visaClaims = getJwtClaim(visaJwt, GA4GH_VISA_V1_CLAIM);
```
zarsky-broad, line 215: I think this will work:
```suggestion
    var visaClaims = getJwtClaim<JSONObject>(visaJwt, GA4GH_VISA_V1_CLAIM);
```

dvoet: nope, but this works
`var visaClaims = this.<JSONObject>getJwtClaim(visaJwt, GA4GH_VISA_V1_CLAIM);`
but I don't really like it

PR #33, opened 2021-07-23
restructured transactions

dvoet:
This started out as making `ProviderService` call `linkedAccountService.saveLinkedAccount` instead of `OidcApiController`. This led to propagating database ids in return types (`saveLinkedAccount` returns `LinkedAccountWithPassportAndVisas` with populated database ids).

Then I discovered that `OidcApiController` did not have a test for `createLink`.

Then I moved the `@Transactional` and `@Rollback` annotations to `BaseTest` because I noticed some tests did not have them and were littering the db and really almost all tests should have them (except those testing fancy transaction dances).

And that was my morning.


Shakespeared (approved):
LGTM

snwessel (approved):
LGTM

PR #34, opened 2021-07-23
Immutables

dvoet:
Here is an alternative to lombok. It is a little more opinionated about required fields, nullable, etc. which is probably a good thing. I looked at it because it might be more compatible with stairway should we need it. It is more traditional code gen than lombok.


snwessel (approved):
LGTM! I'm personally a big fan of all the optionals!

zarsky-broad:
I want to understand this better, can we chat about it next week?

dvoet:
absolutely… this is a proposal, I am still thinking about it too

dvoet:
things I like:
forced use of optionals and null checks - also doable with lombok but lombok defaults to allow nulls
immutable by default - lombok requires fields to be marked as final
matches our usecase more closely - lombok can do more stuff but we don't need it and shouldn't use it

Immutables generates new classes which are the ones we actually use in code which makes code navigation not as good (click through goes to the generated class), however it is more transparent. Lombok alters the bytecode so essentially hides the details of what it does (although you can delombok it). The problem with lombok is that most experienced java coders will say WTF because they distrust magic.

zarsky-broad (approved):

PR #37, opened 2021-07-27
POL-35 delete provider link

dvoet:


service/src/main/java/bio/terra/externalcreds/services/ProviderService.java
```diff
@@ -122,6 +128,52 @@ public LinkedAccountWithPassportAndVisas createLink(
     return linkedAccountService.saveLinkedAccount(extractPassport(userInfo, linkedAccount));
   }
 
+  public void deleteLink(String userId, String providerId) {
```
snwessel, line 131: Should this be a transaction so that we ensure nothing changes with the DB record between our calls to  `getLinkedAccount` and `deleteLinkedAccount`? I know that would leave it hanging for a bit though while we revoke the access token.

service/src/main/java/bio/terra/externalcreds/controllers/OidcApiController.java
```diff
@@ -117,4 +117,11 @@ private LinkInfo getLinkInfoFromLinkedAccount(LinkedAccount linkedAccount) {
     return ResponseEntity.ok(
         getLinkInfoFromLinkedAccount(linkedAccountWithPassportAndVisas.getLinkedAccount()));
   }
+
+  @Override
+  public ResponseEntity<Void> deleteLink(String provider) {
```
snwessel, line 122: Do you think we should create an explicit test for this controller method? There’s not too much going on in it that hasn't already been tested, but I think our approach in the past had been to test all the controller methods regardless, right?

service/src/test/java/bio/terra/externalcreds/services/OAuth2ServiceTest.java
```diff
@@ -60,6 +60,9 @@ void test() {
         oAuth2Service.authorizeWithRefreshToken(
             providerClient, oAuth2AccessTokenResponse.getRefreshToken());
 
+    System.out.println(
+        "refresh token:__________" + tokenResponse.getRefreshToken().getTokenValue());
+
```
Shakespeared, line 65: testing line left in here

service/src/main/java/bio/terra/externalcreds/services/ProviderService.java
```diff
@@ -122,6 +128,52 @@ public LinkedAccountWithPassportAndVisas createLink(
...
+                        .queryParam("client_id", providerInfo.getClientId())
+                        .queryParam("client_secret", providerInfo.getClientSecret())
+                        .build())
+            .retrieve();
+
+    String responseBody =
+        response
+            .onStatus(HttpStatus::isError, clientResponse -> Mono.empty())
+            .bodyToMono(String.class)
+            .block(Duration.of(1000, ChronoUnit.MILLIS));
```
snwessel, line 168: Two quick questions about this part:
1. Is this the correct duration we should wait when calling the revoke endpoint? I had chosen 1 second arbitrarily
2. Should we throw an exception when the revoke endpoint returns an error? My thought had been that we would want to prevent the record from being deleted in the DB whenever we've failed to revoke it. Does that sound right?

Shakespeared (approved):
LGTM overall but seconding Sarah's comments

service/src/main/java/bio/terra/externalcreds/controllers/OidcApiController.java
dvoet: yes

service/src/main/java/bio/terra/externalcreds/services/ProviderService.java
dvoet: We should avoid making external api calls within a transaction.  The worst that can happen here is that a user concurrently deletes and creates their link. If the create happens after the delete reads the refresh token but before it deletes the link then the new refresh token obtained by create process will be lost. Not a big deal.

I also want to make sure that it is known that in general transactions don't give this kind of protection anyway. The serializable isolation level is the thing that would give protection.

service/src/main/java/bio/terra/externalcreds/services/ProviderService.java
dvoet: 1s sounds fine to me. The revoke call should be a best effort attempt and should not prevent the user from deleting their link. So I did not throw an error on failure.

service/src/test/java/bio/terra/externalcreds/services/OAuth2ServiceTest.java
dvoet: I thought this was fine

service/src/main/java/bio/terra/externalcreds/controllers/OidcApiController.java
snwessel: Okay sounds good. I actually already got started on writing a test for this on Friday, so I would be happy to wrap that up this afternoon, and maybe create a `TestUtils` class in the process. Sound reasonable?

Also, if you prefer to merge this now, go ahead and I can create a separate PR for the test.

snwessel (approved):

snwessel:
I just added a test for the controller (which mostly just ensures that it gets the user Id without an error), and moved some of the duplicate private helpers into a TestUtils class. Let me know if anyone has any feedback on that!

service/src/test/java/bio/terra/externalcreds/services/OAuth2ServiceTest.java
Shakespeared: ok! If you think it's fine then I'm fine to leave it

PR #39, opened 2021-07-29
fix jibDockerBuild failure

dvoet:


.github/workflows/build-and-test.yml
```diff
@@ -48,3 +48,6 @@ jobs:
 
       - name: Build and test with Gradle
         run: ./gradlew build
+      - name: Build image locally with jib
+        # build the docker image to make sure it does not error
+        run: './gradlew :service:jibDockerBuild -Djib.console=plain'
```
dvoet, line 53: added this so we can catch the problem before PRs merge

build.gradle
```diff
@@ -1,3 +1,12 @@
+buildscript {
+	dependencies {
+		// this fixes the error
+		// Execution failed for task ':service:jibDockerBuild'.
+		// > 'java.util.stream.Collector com.google.common.collect.ImmutableList.toImmutableList()'
+		classpath 'com.google.guava:guava:30.1.1-jre'
```
dvoet, line 6: this seems to force the use of the right version of guava

Shakespeared (approved):
Sounds good to me

snwessel (approved):

PR #40, opened 2021-07-30
Play

dvoet:


service/src/main/java/bio/terra/externalcreds/dataAccess/GA4GHVisaDAO.java
```diff
@@ -48,7 +49,7 @@ public GA4GHVisa insertVisa(GA4GHVisa visa) {
     var generatedKeyHolder = new GeneratedKeyHolder();
     jdbcTemplate.update(query, namedParameters, generatedKeyHolder);
 
-    return visa.withId(generatedKeyHolder.getKey().intValue());
+    return ImmutableGA4GHVisa.copyOf(visa).withId(generatedKeyHolder.getKey().intValue());
```
snwessel, line 52: My understanding is that the `withId`method by itself already creates a copy (screenshot below). Is the `ImmutableGA4GHVisa.copyOf(...)` method used here because just using `withId` would output the wrong (mutable) object type? Or is there some other reason for using this? I want to make sure I understand so I know when and where to use `copyOf` in the future. 

image

dvoet: the type of `visa` here is `GA4GHVisa`, not `ImmutableGA4GHVisa`. The `GA4GHVisa` interface does not have a `withId` function so we need to copy it to get an instance of `ImmutableGA4GHVisa` (which if you look at the implementation of `copyOf` does not actually copy if it doesn't have to)

snwessel: Okay great, that makes sense! Thanks for the explanation.

snwessel (approved):

PR #41, opened 2021-08-02
get provider passport [POL-46]

Shakespeared:
get a passport from a particular provider from the DB

### DIFF
--- a/service/src/main/java/bio/terra/externalcreds/dataAccess/GA4GHPassportDAO.java
+++ b/service/src/main/java/bio/terra/externalcreds/dataAccess/GA4GHPassportDAO.java
@@ -18,12 +18,10 @@ import org.springframework.stereotype.Repository;
 public class GA4GHPassportDAO {
 
   final NamedParameterJdbcTemplate jdbcTemplate;
-  final LinkedAccountDAO linkedAccountDAO;
 
   public GA4GHPassportDAO(
-      NamedParameterJdbcTemplate jdbcTemplate, LinkedAccountDAO linkedAccountDAO) {
+      NamedParameterJdbcTemplate jdbcTemplate) {
     this.jdbcTemplate = jdbcTemplate;
-    this.linkedAccountDAO = linkedAccountDAO;
   }
 
   /**
--- a/service/src/main/java/bio/terra/externalcreds/dataAccess/GA4GHPassportDAO.java
+++ b/service/src/main/java/bio/terra/externalcreds/dataAccess/GA4GHPassportDAO.java
@@ -18,9 +18,12 @@ import org.springframework.stereotype.Repository;
 public class GA4GHPassportDAO {
 
   final NamedParameterJdbcTemplate jdbcTemplate;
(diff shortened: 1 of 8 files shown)


service/src/main/java/bio/terra/externalcreds/services/LinkedAccountService.java
```diff
@@ -37,6 +39,13 @@ public LinkedAccountService(
     return linkedAccountDAO.getLinkedAccount(userId, providerId);
   }
 
+  @ReadTransaction
+  public Optional<GA4GHPassport> getGA4GHPassport(String userId, String providerId) {
+    var linkedAccount = linkedAccountDAO.getLinkedAccount(userId, providerId);
+    if (linkedAccount.isEmpty()) throw new NotFoundException("Linked account not found");
+    return ga4ghPassportDAO.getPassport(linkedAccount.get().getId().orElseThrow());
```
dvoet: This can be done in a single database query, not 2. It looks like `ga4ghPassportDAO.getPassport(linkedAccountId)` is only called by tests. Perhaps that function should be changed to accept userId and providerId as parameters instead.

service/src/test/java/bio/terra/externalcreds/services/LinkedAccountServiceTest.java
```diff
@@ -44,6 +44,22 @@ void testGetLinkedAccount() {
         savedLinkedAccount.map(ImmutableLinkedAccount::copyOf).get().withId(Optional.empty()));
   }
 
+  @Test
+  void testGetGA4GHPassport() {
```
dvoet, line 48: also test `getGA4GHPassport` when the linked account does not exist and when the linked account does exist but passport does not

dvoet (changes requested):

service/src/main/java/bio/terra/externalcreds/services/LinkedAccountService.java
Shakespeared: I think I got it this time…

service/src/main/java/bio/terra/externalcreds/dataAccess/GA4GHPassportDAO.java
```diff
@@ -18,9 +18,12 @@
 public class GA4GHPassportDAO {
 
   final NamedParameterJdbcTemplate jdbcTemplate;
+  final LinkedAccountDAO linkedAccountDAO;
```
dvoet: don't need this and in general daos should not depend on other daos

dvoet (approved):
remove some leftovers then good to merge

dvoet (approved):

service/src/main/java/bio/terra/externalcreds/dataAccess/GA4GHPassportDAO.java
Shakespeared: oh oops that was left over from a previous thing I was trying, thanks!

snwessel (approved):

PR #43, opened 2021-08-05
More optionals

dvoet:
changed `providerClientCache.getProviderClient()` and `providerService.createLink()` to return optional when given provider does not exist

### DIFF
--- a/service/src/main/java/bio/terra/externalcreds/controllers/OidcApiController.java
+++ b/service/src/main/java/bio/terra/externalcreds/controllers/OidcApiController.java
@@ -104,8 +104,9 @@ public class OidcApiController implements OidcApi {
         providerService.createLink(
             provider, userId, oauthcode, redirectUri, Set.copyOf(scopes), state);
 
-    return ResponseEntity.ok(
-        getLinkInfoFromLinkedAccount(linkedAccountWithPassportAndVisas.getLinkedAccount()));
+    return ResponseEntity.of(
+        linkedAccountWithPassportAndVisas.map(
+            x -> getLinkInfoFromLinkedAccount(x.getLinkedAccount())));
   }
 
   @Override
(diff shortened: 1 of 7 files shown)


service/src/main/java/bio/terra/externalcreds/controllers/OidcApiController.java
```diff
@@ -104,8 +104,9 @@ private LinkInfo getLinkInfoFromLinkedAccount(LinkedAccount linkedAccount) {
         providerService.createLink(
             provider, userId, oauthcode, redirectUri, Set.copyOf(scopes), state);
 
-    return ResponseEntity.ok(
-        getLinkInfoFromLinkedAccount(linkedAccountWithPassportAndVisas.getLinkedAccount()));
+    return ResponseEntity.of(
+        linkedAccountWithPassportAndVisas.map(
+            x -> getLinkInfoFromLinkedAccount(x.getLinkedAccount())));
```
dvoet, line 109: this could also be written as
```
    return ResponseEntity.of(
        linkedAccountWithPassportAndVisas
            .map(LinkedAccountWithPassportAndVisas::getLinkedAccount)
            .map(this::getLinkInfoFromLinkedAccount));
```
Which do you like better?

Shakespeared: Either works for me!

Shakespeared (approved):

service/src/main/java/bio/terra/externalcreds/controllers/OidcApiController.java
snwessel: I'm more familiar with the syntax of what you have now, but either would be fine with me!

snwessel (approved):

PR #47, opened 2021-08-10
Test Runner Initial Setup [POL-48]

snwessel:


.github/workflows/pr-integration.yml
```diff
@@ -0,0 +1,76 @@
...
+          psql -h localhost -U postgres -f ./common/postgres-init.sql
+
+      - name: Render GitHub Secrets
+        id: github-secrets
+        shell: bash
+        env:
+          # Copied to GH from vault location: "secret/dsde/firecloud/dev/common/firecloud-account.json"
+          DEV_FIRECLOUD_ACCOUNT: ${{ secrets.DEV_FIRECLOUD_ACCOUNT }}
+        run:
+          echo "$DEV_FIRECLOUD_ACCOUNT" >"./integration/src/main/resources/rendered/user-delegated-sa.json"
```
snwessel: I confirmed that this doesn't log the secret when it runs: 
image

integration/README.md
```diff
@@ -0,0 +1,25 @@
...
+3. Debug the test locally using the instructions below. 
+
+
+### Running Integration Tests
+
+The test runner task `runTest` can be used to launch individual tests or entire test suites. 
+
+To run the tests locally:
+
+1. Run the `ExternalCredsApplication` (in IntelliJ).
```
dvoet: link to DEVELOPMENT.md in root for instructions

.github/workflows/pr-integration.yml
snwessel: The value of this secret isn't actually used yet since the `ServiceStatus` test doesn't require auth or interact with other services, but we'll need it in the near future to do these things. Also test runner has an issue where it requires this be defined even if it isn't using it, so I went ahead and set it up.

integration/src/main/java/scripts/testscripts/ServiceStatus.java
```diff
@@ -0,0 +1,38 @@
...
+
+    @Override
+    public void userJourney(TestUserSpecification testUser) throws Exception {
+        if (delay.getSeconds() > 0) TimeUnit.SECONDS.sleep(delay.getSeconds());
+
+         log.info("Checking service status endpoint now.");
+         var apiClient = ClientTestUtils.getClientWithoutAccessToken(server);
+         var publicApi = new PublicApi(apiClient);
+         publicApi.getStatus();
+         var httpCode = publicApi.getApiClient().getStatusCode();
```
dvoet: should this check the response code?

integration/src/main/resources/servers/ecm-local.json
```diff
@@ -0,0 +1,12 @@
+{
+  "name": "ecm-local",
+  "description": "ECM Local env.",
+
+  "samUri": "https://sam.dsde-dev.broadinstitute.org",
+  "workspaceManagerUri": "http://localhost:8080",
```
dvoet: we should not need a `workspaceManagerUri` setting

dvoet (approved):
a couple nits

Shakespeared (approved):
LGTM

PR #49, opened 2021-08-13
POL-51 background tasks in ECM

dvoet:
https://broadworkbench.atlassian.net/browse/POL-51


service/src/main/java/bio/terra/externalcreds/ExternalCredsWebApplication.java
```diff
@@ -0,0 +1,28 @@
...
+@SpringBootConfiguration
+@EnableAutoConfiguration
+@ComponentScan(
+    basePackages = {
+      "bio.terra.externalcreds",
+      "bio.terra.common.logging",
+      "bio.terra.common.retry.transaction"
+    },
+    excludeFilters = @Filter(type = FilterType.ANNOTATION, classes = SpringBootConfiguration.class))
+public class ExternalCredsWebApplication {
```
dvoet, line 21: the is a rename from `ExternalCredsApplication` except it no longer uses `@SpringBootApplication` because that does not give control over exclusions

service/src/main/resources/application.yml
```diff
@@ -27,10 +27,23 @@ spring:
...
           must-revalidate: true
         use-last-modified: false
+
+---
+spring:
+  config:
+    activate:
+      on-profile: cron-app
+  main:
+    web-application-type: none
```
dvoet: this tells spring not to start a web server

service/src/main/java/bio/terra/externalcreds/ExternalCredsCronApplication.java
```diff
@@ -0,0 +1,48 @@
...
+    basePackages = {
+      "bio.terra.externalcreds",
+      "bio.terra.common.logging",
+      "bio.terra.common.retry.transaction"
+    },
+    excludeFilters =
+        @Filter(
+            type = FilterType.ANNOTATION,
+            classes = {Controller.class, SpringBootConfiguration.class}))
+public class ExternalCredsCronApplication implements ApplicationRunner {
```
dvoet, line 29: https://www.baeldung.com/running-setup-logic-on-startup-in-spring#7-spring-boot-applicationrunner

.github/workflows/publish-branch-image.yml
```diff
@@ -0,0 +1,51 @@
+name: Publish branch image
```
Shakespeared, line 1: @snwessel and I noticed this is very similar to the `publish` workflow that already exists, can you point out what this one adds that requires a separate workflow, as opposed to being part of the existing workflow?

Shakespeared (approved):

.github/workflows/publish-branch-image.yml
dvoet: The publish workflow runs automatically on push to dev and at the end triggers a deploy. This workflow runs on demand on branches and does not trigger a deploy. @zarsky-broad offered to clean this up :)

snwessel (approved):

PR #51, opened 2021-08-17
Refresh expiring visas [POL-49]

snwessel:
It would be helpful to get some feedback on the error handling part of this. We had talked about deleting the linked account when we receive a 400 status code from the provider. However, Kai and I spent a substantial amount of time looking into this and haven't found an easy way to get the code. Isaac and I talked talked through this today and came up with a possible solution (implemented below): We keep track of any linked accounts which encounter errors when their passports are being refreshed. Then, assuming we go with the java-based not-cron approach you described this morning, to call this functionality, it would be easy to handle retries there. Once the expiration time has actually passed, we can give up on retries and delete the linked account (or just the passport). Let me know if you have any thoughts on or objections to that approach.

Also, it's worth noting that we refactored a few things which make up a large portion of the changes below - mostly just moving some existing methods from the `ProviderService` into new `JwtUtils` and `PassportService` classes.


service/src/main/java/bio/terra/externalcreds/services/PassportService.java
```diff
@@ -0,0 +1,73 @@
...
+  }
+
+  @ReadTransaction
+  public Optional<GA4GHPassport> getGA4GHPassport(String userId, String providerId) {
+    return passportDAO.getPassport(userId, providerId);
+  }
+
+  public void authAndRefreshPassport(LinkedAccount linkedAccount) {
+    if (!linkedAccount.getExpires().after(Timestamp.from(Instant.now()))) {
+      passportDAO.deletePassport(linkedAccount.getId().get());
```
dvoet: This is happening outside of a transaction.

service/src/main/java/bio/terra/externalcreds/services/ProviderService.java
```diff
@@ -41,33 +21,53 @@
...
-  public static final String GA4GH_VISA_V1_CLAIM = "ga4gh_visa_v1";
-  public static final String VISA_TYPE_CLAIM = "type";
-  public static final String JKU_HEADER = "jku";
   public static final String EXTERNAL_USERID_ATTR = "email";
 
   private final ExternalCredsConfig externalCredsConfig;
   private final ProviderClientCache providerClientCache;
   private final OAuth2Service oAuth2Service;
   private final LinkedAccountService linkedAccountService;
+  private final LinkedAccountDAO linkedAccountDAO;
```
dvoet: `ProviderService` does not use any transactional annotations so should not access DAOs directly. Pipe calls to `linkedAccountDAO` through `linkedAccountService` with proper transactional annotations in `LinkedAccountService`

service/src/main/java/bio/terra/externalcreds/services/PassportService.java
```diff
@@ -0,0 +1,73 @@
...
+import lombok.extern.slf4j.Slf4j;
+import org.springframework.security.oauth2.core.OAuth2RefreshToken;
+import org.springframework.security.oauth2.core.endpoint.OAuth2AccessTokenResponse;
+import org.springframework.stereotype.Service;
+
+@Service
+@Slf4j
+public class PassportService {
+
+  private final GA4GHPassportDAO passportDAO;
```
dvoet, line 14: `PassportService` does not use any transactional annotations so should not access DAOs directly. Pipe calls to `linkedAccountDAO` through `linkedAccountService` with proper transactional annotations in `LinkedAccountService`

I am also not convinced `PassportService` should exist.

service/src/main/java/bio/terra/externalcreds/config/ExternalCredsConfigInterface.java
```diff
@@ -16,4 +16,6 @@
   VersionProperties getVersion();
 
   String getSamBasePath();
+
+  int getVisaAndPassportRefreshInterval();
```
dvoet: this should return a `Duration`

dvoet (changes requested):

service/src/main/java/bio/terra/externalcreds/services/PassportService.java
```diff
@@ -0,0 +1,73 @@
...
+            Optional.ofNullable(
+                oAuth2Service.authorizeWithRefreshToken(
+                    clientRegistration,
+                    new OAuth2RefreshToken(linkedAccount.getRefreshToken(), null)));
+      } catch (Exception e) {
+        // An IllegalArgumentException is thrown when the issuer can not be reached
+        // An OAuth2AuthorizationException is thrown when the provider returns a non-200 response
+        // TODO: decide how to handle different exceptions differently
+        //   (upon failure to auth, consider deleting the linked account?)
+        log.error("Encountered an error while getting a new refresh token: {}", e.getMessage());
```
dvoet: this logs just the exception message and looses all other information like exception class and stack trace. You should almost never do this. Instead
```
log.error("Encountered an error while getting a new refresh token", e);
```

service/src/main/java/bio/terra/externalcreds/services/PassportService.java
```diff
@@ -0,0 +1,73 @@
...
+            "Encountered an error while getting a new refresh token", e);
+      }
+      if (accessTokenResponse.isEmpty()) {
+        linkedAccountService.deleteLinkedAccount(
+            linkedAccount.getUserId(), linkedAccount.getProviderId());
+      } else {
+        var userInfo =
+            oAuth2Service.getUserInfo(
+                clientRegistration, accessTokenResponse.get().getAccessToken());
+        linkedAccountService.upsertLinkedAccount(JwtUtils.extractPassport(userInfo, linkedAccount));
```
dvoet: this does not save the new refresh token if it exists

service/src/main/java/bio/terra/externalcreds/dataAccess/LinkedAccountDAO.java
```diff
@@ -32,10 +34,23 @@ public LinkedAccountDAO(NamedParameterJdbcTemplate jdbcTemplate) {
...
 
+  public List<LinkedAccount> getExpiringLinkedAccounts(Timestamp expirationCutoff) {
+    var namedParameters = new MapSqlParameterSource("expirationCutoff", expirationCutoff);
+    var query =
+        "SELECT DISTINCT la.* FROM linked_account la"
+            + " JOIN ga4gh_passport passport"
+            + " ON passport.linked_account_id = la.id"
+            + " LEFT JOIN ga4gh_visa visa"
+            + " ON visa.passport_id = passport.id"
+            + " WHERE passport.expires <= :expirationCutoff"
```
nmalfroy: unless you clear out the tables, this will include expired passports and visas too no?  You may want a way to filter those out

service/src/main/java/bio/terra/externalcreds/models/LinkedAccount.java
```diff
@@ -18,5 +18,10 @@
 
   String getExternalUserId();
 
+  @Value.Default
+  default boolean isAuthenticated() {
+    return true;
```
nmalfroy: Not sure of the intent, but should default for authentication be false?

service/src/main/java/bio/terra/externalcreds/services/JwtUtils.java
```diff
@@ -0,0 +1,147 @@
...
+
+  private final ExternalCredsConfig externalCredsConfig;
+
+  public JwtUtils(ExternalCredsConfig externalCredsConfig) {
+    this.externalCredsConfig = externalCredsConfig;
+  }
+
+  public LinkedAccountWithPassportAndVisas enrichAccountWithPassportAndVisas(
+      LinkedAccount linkedAccount, OAuth2User userInfo) {
+    var passportJwtString = userInfo.<String>getAttribute(PASSPORT_JWT_V11_CLAIM);
```
nmalfroy: Style thing but it seems like using vars and then needing to specify Type doesn't buy you anything and just makes the code look kinda weird.  Seems like it's just as easy (and arguably more readable to do:
```suggestion
    String passportJwtString = userInfo.getAttribute(PASSPORT_JWT_V11_CLAIM);
```

service/src/main/java/bio/terra/externalcreds/services/JwtUtils.java
```diff
@@ -0,0 +1,147 @@
...
+    if (passportJwtString != null) {
+      var passportJwt = decodeJwt(passportJwtString);
+
+      var visaJwtStrings =
```
nmalfroy: Same comment about the `var` + generics pattern

service/src/main/java/bio/terra/externalcreds/services/JwtUtils.java
```diff
@@ -0,0 +1,147 @@
...
+          Objects.requireNonNullElse(
+              passportJwt.getClaimAsStringList(GA4GH_PASSPORT_V1_CLAIM),
+              Collections.<String>emptyList());
+
+      var visas =
+          visaJwtStrings.stream().map(s -> buildVisa(decodeJwt(s))).collect(Collectors.toList());
```
nmalfroy: When I can I like to chain these calls, e.g. 
```suggestion
          visaJwtStrings.stream().map(this::decodeJwt).map(this::buildVisa).collect(Collectors.toList());
```

service/src/main/java/bio/terra/externalcreds/services/JwtUtils.java
```diff
@@ -0,0 +1,147 @@
...
+      return new LinkedAccountWithPassportAndVisas.Builder().linkedAccount(linkedAccount).build();
+    }
+  }
+
+  private static GA4GHPassport buildPassport(Jwt passportJwt) {
+    var passportExpiresAt = getJwtExpires(passportJwt);
+
+    return new GA4GHPassport.Builder()
+        .jwt(passportJwt.getTokenValue())
+        .expires(passportExpiresAt)
```
nmalfroy, line 75: meta point: this may be getting ahead but you may want a lower bound time too

service/src/main/java/bio/terra/externalcreds/services/JwtUtils.java
```diff
@@ -0,0 +1,147 @@
...
+  private static Timestamp getJwtExpires(Jwt decodedPassportJwt) {
+    var expiresAt = decodedPassportJwt.getExpiresAt();
+    if (expiresAt == null) {
+      throw new InvalidJwtException("jwt missing expires (exp) claim");
+    }
+    return new Timestamp(expiresAt.toEpochMilli());
+  }
+
+  private static <T> T getJwtClaim(Jwt jwt, String claimName) {
+    T claim = jwt.getClaim(claimName);
```
nmalfroy, line 104: Can do this in a single call
```suggestion
    return Optional.ofNullable(jwt.getClaim(claimName)).orElseThrow(() -> new InvalidJwtException(String.format("jwt missing claim [%s]", claimName)));
```

service/src/main/java/bio/terra/externalcreds/services/JwtUtils.java
```diff
@@ -0,0 +1,147 @@
...
+  Jwt decodeJwt(String jwtString) {
+    try {
+      // first we need to get the issuer from the jwt, the issuer is needed to validate
+      var jwt = JWTParser.parse(jwtString);
+      var issuer = jwt.getJWTClaimsSet().getIssuer();
+      if (issuer == null) {
+        throw new InvalidJwtException("jwt missing issuer (iss) claim");
+      }
+      var jkuOption = Optional.ofNullable(((JWSHeader) jwt.getHeader()).getJWKURL());
+      if (jkuOption.isPresent()) {
```
nmalfroy, line 127: FWIW, you can simplify this a bit with:
```suggestion
      return jkuOption.map(jku ->
      if (jkuOption.isPresent()) {
        if (externalCredsConfig.getAllowedJwksUris().contains(jku)) {
          return ExternalCredsJwtDecoders.fromJku(jku).decode(jwtString);
        } else {
          throw new InvalidJwtException(
              String.format("URI [%s] specified by jku header not on allowed list", jku));
        }).orElseGet(() -> JwtDecoders.fromIssuerLocation(issuer).decode(jwtString))
```

service/src/main/java/bio/terra/externalcreds/services/PassportService.java
nmalfroy: I'm curious: is this because it's currently a 1:1?  This comment might be outdated since right now it looks like there are transactional annotations

service/src/main/java/bio/terra/externalcreds/services/ProviderService.java
```diff
@@ -144,7 +133,8 @@ private LinkedAccountWithPassportAndVisas createLinkInternal(
             .refreshToken(refreshToken.getTokenValue())
             .build();
 
-    return linkedAccountService.saveLinkedAccount(extractPassport(userInfo, linkedAccount));
+    return linkedAccountService.upsertLinkedAccountWithPassportAndVisas(
```
nmalfroy, line 151: Should this method be transactional since it looks like it's reaching across services?

service/src/main/java/bio/terra/externalcreds/services/ProviderService.java
```diff
@@ -164,6 +154,47 @@ public void deleteLink(String userId, String providerId) {
...
+    if (linkedAccount.getExpires().before(Timestamp.from(Instant.now()))) {
+      passportService.deletePassport(linkedAccount.getId().get());
+      linkedAccountService.updateLinkAuthenticationStatus(linkedAccount.getId().get(), false);
+    } else {
+      try {
+        var clientRegistration =
+            providerClientCache.getProviderClient(linkedAccount.getProviderId());
+        var accessTokenResponse =
+            oAuth2Service.authorizeWithRefreshToken(
+                clientRegistration.orElseThrow(),
```
nmalfroy: Or else throw what?  This is kinda my issue with `var` notation...it's hard to see what class this is for so you can't really tell right away

service/src/main/java/bio/terra/externalcreds/services/ProviderService.java
```diff
@@ -164,6 +154,47 @@ public void deleteLink(String userId, String providerId) {
...
+        // correctly
+        // save the linked account with the new refresh token and extracted passport
+        var linkedAccountWithRefreshToken =
+            linkedAccount.withRefreshToken(accessTokenResponse.getRefreshToken().getTokenValue());
+        linkedAccountService.upsertLinkedAccountWithPassportAndVisas(
+            jwtUtils.enrichAccountWithPassportAndVisas(linkedAccountWithRefreshToken, userInfo));
+
+      } catch (IllegalArgumentException iae) {
+        throw new ExternalCredsException(
+            "Could not contact issuer for provider " + linkedAccount.getProviderId(), iae);
```
nmalfroy: Should use `String.format`

service/src/main/java/bio/terra/externalcreds/services/ProviderService.java
```diff
@@ -164,6 +154,47 @@ public void deleteLink(String userId, String providerId) {
...
+      } catch (OAuth2AuthorizationException oauthEx) {
+        // if the cause is a 4xx response, delete the passport
+        if (oauthEx.getCause() instanceof HttpClientErrorException) {
+          passportService.deletePassport(linkedAccount.getId().get());
+          linkedAccountService.updateLinkAuthenticationStatus(linkedAccount.getId().get(), false);
+        } else {
+          // log and try again later
+          throw new ExternalCredsException("Failed to refresh passport: ", oauthEx);
```
nmalfroy, line 221: Should use `String.format`

nmalfroy (commented):
Cool! I did a super quick pass.  Everything seems pretty reasonable but I'm missing chunks of context I'm sure.  A lot of my comments are more stylistic.  I do have a question about the DB and expired passports visas being returned in the expiring list

service/src/main/java/bio/terra/externalcreds/services/JwtUtils.java
snwessel: That's a good point! I think in this case especially it would make more sense to just declare the type.

service/src/main/java/bio/terra/externalcreds/models/LinkedAccount.java
snwessel: Good point. Right now, linked accounts are always authenticated when they are created - the only possible "unauthenticated" linked account is one that has become invalid somehow (like if there were errors refreshing it in time). I think there's an argument to be had for not using a default at all, but maybe we should talk that over as a team.

service/src/main/java/bio/terra/externalcreds/dataAccess/LinkedAccountDAO.java
Shakespeared: It is possible for passports to become expired, but in that case we'd want to keep checking them, so I think this behavior is ok. When the linked account expires, associated passports will be deleted (and we try to refresh passports before they expire) so they won't be hanging around.

service/src/main/java/bio/terra/externalcreds/services/ProviderService.java
Shakespeared: The methods this calls are transactional, so this should be fine

service/src/main/java/bio/terra/externalcreds/services/ProviderService.java
Shakespeared: That's a good point! If we change this we'd want to do it across the code base, but since it's not specific to this PR I think we'll want to discuss as a team and decide so I'm going to leave it for now.

service/src/main/java/bio/terra/externalcreds/services/ProviderService.java
Shakespeared: I believe that formatting the exception causes the stacktrace to be lost (see comment: https://github.com/DataBiosphere/terra-external-credentials-manager/pull/51#discussion_r691470490)

service/src/main/java/bio/terra/externalcreds/services/PassportService.java
snwessel: Yeah, we've updated the `PassportService` class a lot since Doug's original comment was made, and it now has transactional annotations. We decided to keep the class because it would be a more intuitive place to look for passport-related functionality than `LinkedAccountService`.

service/src/main/java/bio/terra/externalcreds/services/JwtUtils.java
snwessel: I'm not sure I follow what you mean by the lower bound. Are you suggesting adding something like a `created_at` `refresh_after` timestamp to the passport model?

service/src/main/java/bio/terra/externalcreds/services/JwtUtils.java
snwessel: Thank you for the suggestion! For some reason, the map function doesn't like that an exception could be thrown inside of it and caught outside of it, so we might just leave this part as it currently is.

snwessel:
@nmalfroy I think we've addressed all your comments. If you have the chance to give it a final thumb or give us any other feedback, that would be great! 

Thank you again for reviewing!

service/src/main/java/bio/terra/externalcreds/services/JwtUtils.java
nmalfroy: Sorry yeah exactly.  It's a corner case but I've seen a lot of similar systems implement that start effective date.  Seems like adding more restriction is likely safer?

service/src/main/java/bio/terra/externalcreds/services/JwtUtils.java
nmalfroy: fair enough!

service/src/main/java/bio/terra/externalcreds/services/ProviderService.java
nmalfroy: Got it.  Just didn't know if the calls needed to be in a global transaction

nmalfroy: (I ask because we ran into similar issues in TDR)

service/src/main/java/bio/terra/externalcreds/services/ProviderService.java
nmalfroy: ah you're totally right!  Sorry, I had just made a comment above and thought it applied here but you're right, it doesn't

service/src/main/java/bio/terra/externalcreds/services/JwtUtils.java
snwessel: That sounds like a good thing to keep in mind, but may be an architectural change that is larger than the scope of this PR. Doug is out on vacation right now, but we can bring this up with him once he's back and think about implementing it.

service/src/main/java/bio/terra/externalcreds/services/ProviderService.java
snwessel: That's a good point - we wouldn't want this method to fail halfway through and not roll back all the way. Just to clarify, are you suggesting that we use [nested transactions](http://mvpjava.com/spring-nested-transactions/)? I haven't worked with them before, but it seems like they are enabled by default. 

If you have worked with them before, would you be able to confirm that the only thing we'd need to do is add another `@Transactional` annotation to the top of this method?

Thanks again!

service/src/main/java/bio/terra/externalcreds/services/ProviderService.java
```diff
@@ -4,69 +4,72 @@
...
     this.linkedAccountService = linkedAccountService;
+    this.passportService = passportService;
+    this.jwtUtils = jwtUtils;
   }
 
   public Set<String> getProviderList() {
     return Collections.unmodifiableSet(externalCredsConfig.getProviders().keySet());
   }
 
+  public void refreshExpiringPassports() {
```
snwessel, line 59: Any thoughts on whether/how to test this function? 

We tried using `Mockito.verify` to check the number of times that `ProviderService.authAndRefreshPassport` was called, but realized we can't do that because `providerService` would have to be a mock.

service/src/main/java/bio/terra/externalcreds/dataAccess/LinkedAccountDAO.java
```diff
@@ -32,10 +34,23 @@ public LinkedAccountDAO(NamedParameterJdbcTemplate jdbcTemplate) {
...
+    return jdbcTemplate.query(query, namedParameters, new LinkedAccountRowMapper());
+  }
+
   public LinkedAccount upsertLinkedAccount(LinkedAccount linkedAccount) {
     var query =
-        "INSERT INTO linked_account (user_id, provider_id, refresh_token, expires, external_user_id)"
-            + " VALUES (:userId, :providerId, :refreshToken, :expires, :externalUserId)"
+        "INSERT INTO linked_account (user_id, provider_id, refresh_token, expires, external_user_id, is_authenticated)"
+            + " VALUES (:userId, :providerId, :refreshToken, :expires, :externalUserId, :isAuthenticated)"
             + " ON CONFLICT (user_id, provider_id) DO UPDATE SET"
```
dvoet, line 55: I think `isAuthenticated` needs to be set in the ON CONFLICT clause too

service/src/main/java/bio/terra/externalcreds/services/JwtUtils.java
dvoet: the jwt does have an issued at (iat) date… we are not pulling it out and saving it separately because we don't have any functionality that cares about it at the moment

service/src/main/java/bio/terra/externalcreds/services/LinkedAccountService.java
```diff
@@ -58,7 +54,17 @@ public boolean deleteLinkedAccount(String userId, String providerId) {
     return linkedAccountDAO.deleteLinkedAccountIfExists(userId, providerId);
   }
 
-  private LinkedAccountWithPassportAndVisas savePassportIfExists(
+  @WriteTransaction
```
dvoet:
```suggestion
  @ReadTransaction
```

service/src/main/java/bio/terra/externalcreds/services/ProviderService.java
dvoet: Did you try [Mockito spy](https://www.baeldung.com/mockito-spy)? But you should be able to check that this updates the database correctly, no?

service/src/main/java/bio/terra/externalcreds/services/ProviderService.java
dvoet: I think the `orElseThrow` should be part of the line above and throw a more descriptive exception

service/src/main/java/bio/terra/externalcreds/dataAccess/LinkedAccountDAO.java
dvoet: should this exclude records where `la.is_authenticated = false`

service/src/main/java/bio/terra/externalcreds/services/ProviderService.java
```diff
@@ -164,6 +169,47 @@ public void deleteLink(String userId, String providerId) {
...
+      linkedAccountService.updateLinkAuthenticationStatus(linkedAccount.getId().get(), false);
+    } else {
+      try {
+        var clientRegistration =
+            providerClientCache.getProviderClient(linkedAccount.getProviderId());
+        var accessTokenResponse =
+            oAuth2Service.authorizeWithRefreshToken(
+                clientRegistration.orElseThrow(),
+                new OAuth2RefreshToken(linkedAccount.getRefreshToken(), null));
+        var userInfo =
```
dvoet: If a new refresh token is returned in the `accessTokenResponse` it should be saved here, before the call to `getUserInfo` because if that call fails we lose the new refresh token and the link becomes unauthenticated because the original refresh token becomes invalid.

service/src/main/java/bio/terra/externalcreds/dataAccess/LinkedAccountDAO.java
```diff
@@ -71,6 +87,17 @@ public boolean deleteLinkedAccountIfExists(String userId, String providerId) {
...
+    var query =
+        "UPDATE linked_account SET is_authenticated = :isAuthenticated WHERE id = :linkedAccountId";
+    var namedParameters =
+        new MapSqlParameterSource()
+            .addValue("linkedAccountId", linkedAccountId)
+            .addValue("isAuthenticated", isAuthenticated);
+
+    return jdbcTemplate.update(query, namedParameters) > 0;
+  }
+
```
dvoet: I am torn about this function. On the one hand, this does just what we want. On the other hand, `upsertLinkedAccount` can do this too so this is extra code to maintain.

service/src/main/java/bio/terra/externalcreds/services/ProviderService.java
```diff
@@ -164,6 +169,47 @@ public void deleteLink(String userId, String providerId) {
     linkedAccountService.deleteLinkedAccount(userId, providerId);
   }
 
+  @VisibleForTesting
+  void authAndRefreshPassport(LinkedAccount linkedAccount) {
+    if (linkedAccount.getExpires().before(Timestamp.from(Instant.now()))) {
+      passportService.deletePassport(linkedAccount.getId().get());
+      linkedAccountService.updateLinkAuthenticationStatus(linkedAccount.getId().get(), false);
```
dvoet:
```suggestion
      var linkedAccountId = linkedAccount.getId().orElseThrow(() -> new ExternalCredsException("linked account id missing"));
      passportService.deletePassport(linkedAccountId);
      linkedAccountService.updateLinkAuthenticationStatus(linkedAccountId, false);
```

dvoet (changes requested):

service/src/main/java/bio/terra/externalcreds/dataAccess/LinkedAccountDAO.java
snwessel: Good point, thank you! I just added it in.

service/src/main/java/bio/terra/externalcreds/services/ProviderService.java
snwessel: Thank you! That totally solved the problem. 

We were trying to avoid doing the whole database check because that would involve mocking quite a few things to get `authAndRefreshPassport` to run without error. Also it seemed like better practice to keep this test as more of a "unit" which doesn't re-test any functionality that has already been well tested.

service/src/main/java/bio/terra/externalcreds/services/LinkedAccountService.java
snwessel: Good catch, thank you! Fixed now.

service/src/main/java/bio/terra/externalcreds/services/ProviderService.java
snwessel: Sounds good! Just updated it.

service/src/main/java/bio/terra/externalcreds/dataAccess/LinkedAccountDAO.java
snwessel: Yes! Good point! In theory, any linked accounts with `la.is_authenticated = false` shouldn't have passports (and won't be returned by the query), but it's definitely better to check just in case.

This is fixed now.

service/src/main/java/bio/terra/externalcreds/dataAccess/LinkedAccountDAO.java
snwessel: Good point - I think using `upsertLinkedAccount` would be simpler. I just added `upsertLinkedAccount` to our `LinkedAccountService` (it was already in the DAO) and got rid of `updateLinkAuthenticationStatus`. That simplified a lot and also made it easier to address the comment you made above.

service/src/main/java/bio/terra/externalcreds/services/ProviderService.java
snwessel: Good point, thanks! Just updated that.

service/src/main/java/bio/terra/externalcreds/services/ProviderService.java
snwessel: Thanks, just updated it!

snwessel:
@dvoet thanks for reviewing! 

I think I was able to address all of your comments i[n this commit](https://github.com/DataBiosphere/terra-external-credentials-manager/pull/51/commits/daba900aee7dc5110c40ba13338c85ace626ec11). Let us know if there is anything else you want us to update. 

Also, let us know what you think of [muscles’s comment about nested transactions](https://github.com/DataBiosphere/terra-external-credentials-manager/pull/51#discussion_r700527119). The [documentation](http://mvpjava.com/spring-nested-transactions/) I saw on it seemed to imply that adding a higher-level transaction annotation wouldn’t do much (because spring would just defer to the lower-level annotations). Have you worked with them at all before? If so, any thoughts on the matter?

service/src/main/java/bio/terra/externalcreds/services/ProviderService.java
```diff
@@ -164,6 +169,64 @@ public void deleteLink(String userId, String providerId) {
...
 
+  @VisibleForTesting
+  void authAndRefreshPassport(LinkedAccount linkedAccount) {
+    if (linkedAccount.getExpires().before(Timestamp.from(Instant.now()))) {
+      var linkedAccountId =
+          linkedAccount
+              .getId()
+              .orElseThrow(() -> new ExternalCredsException("linked account id missing"));
+      passportService.deletePassport(linkedAccountId);
+      linkedAccountService.upsertLinkedAccount(linkedAccount.withIsAuthenticated(false));
```
dvoet:
```suggestion
      linkedAccountService.upsertLinkedAccountWithPassportAndVisas(
          new LinkedAccountWithPassportAndVisas.Builder()
              .linkedAccount(linkedAccount.withIsAuthenticated(false))
              .passport(Optional.empty()) // explicitly set to empty to be clear about intent
              .build());
```
this is nice because the single call is transactional

service/src/main/java/bio/terra/externalcreds/services/ProviderService.java
```diff
@@ -172,23 +172,36 @@ public void deleteLink(String userId, String providerId) {
...
-            oAuth2Service.getUserInfo(
-                clientRegistration.orElseThrow(), accessTokenResponse.getAccessToken());
+                clientRegistration, new OAuth2RefreshToken(linkedAccount.getRefreshToken(), null));
 
         // save the linked account with the new refresh token and extracted passport
         var linkedAccountWithRefreshToken =
-            linkedAccount.withRefreshToken(accessTokenResponse.getRefreshToken().getTokenValue());
+            linkedAccountService.upsertLinkedAccount(
+                linkedAccount.withRefreshToken(
+                    accessTokenResponse.getRefreshToken().getTokenValue()));
```
dvoet: we should only do this if `accessTokenResponse.getRefreshToken()` is not null

service/src/main/java/bio/terra/externalcreds/services/ProviderService.java
```diff
@@ -200,8 +213,12 @@ void authAndRefreshPassport(LinkedAccount linkedAccount) {
...
         // if the cause is a 4xx response, delete the passport
         if (oauthEx.getCause() instanceof HttpClientErrorException) {
-          passportService.deletePassport(linkedAccount.getId().get());
-          linkedAccountService.updateLinkAuthenticationStatus(linkedAccount.getId().get(), false);
+          var linkedAccountId =
+              linkedAccount
+                  .getId()
+                  .orElseThrow(() -> new ExternalCredsException("linked account id missing"));
+          passportService.deletePassport(linkedAccountId);
+          linkedAccountService.upsertLinkedAccount(linkedAccount.withIsAuthenticated(false));
```
dvoet: do the same as line 25, even pull out a function. The we can remove the passportService.deletePassport function

service/src/main/java/bio/terra/externalcreds/services/ProviderService.java
dvoet: I think the code as written is good. Nested transactions are not a thing we want.

It is ok if this function fails part way through, there would be nothing to roll back.

dvoet (approved):

PR #53, opened 2021-08-24
QA-1554: Test Runner upgrade to support optional URI properties and add a new property for policy manager URI.

ichengchang:
These files are updated to support latest version of TestRunner (0.0.8-SNAPSHOT).


snwessel (approved):
Looks good to me! I'm not sure why the test failed on after the last commit, but I'd be happy t o merge it into my branch and debug from there.

PR #55, opened 2021-08-25
Tidier style of immutables

zarsky-broad:


Shakespeared (approved):
👍

snwessel (approved):
👍

dvoet:
@zarsky-broad I am not thrilled with the changes in this PR. It smells funny to me that that the model classes are dependent on the generated classes. Adding the `class Builder` to each model interface also looks like it is trying to hide Immutables but I don't think we want that.

PR #61, opened 2021-09-08
Fix nondeterminism

snwessel:
It looks like this test had been failing randomly (and infrequently) when the linkedAccount's timestamp ended in a zero. Basically, one style of `Timestamp`->`OffsetDateTime`->`String` conversion prints with the final zero, while the other prints without. Kai and I confirmed that both formats are fine (both can be converted back to an `OffsetDateTime` without issue). 

This fix is just to avoid using a random timestamp in the test. 

Screenshot of the failed test output: 
image


dvoet:
I merged the other solution

PR #62, opened 2021-09-09
verify using json serializer instead of hand rolled json

dvoet:
I hope this fixes the intermittent test failures


snwessel (approved):
Looks great! And seems like a much more elegant way of fixing it.

PR #64, opened 2021-09-13
Validate Token type Visas [POL-50]

Shakespeared:
- add config to set default visa validation schedule (currently every hour)
- check for visas that have not been validated within an hour
- validate those visas using the provider validation API endpoint
- for any visas that are not valid, refresh the associated passport with the provider


service/src/main/java/bio/terra/externalcreds/dataAccess/GA4GHPassportDAO.java
```diff
@@ -67,6 +76,25 @@ public GA4GHPassport insertPassport(GA4GHPassport passport) {
             jdbcTemplate.query(query, namedParameters, new GA4GHPassportRowMapper())));
   }
 
+  public List<PassportVerificationDetails> getPassportsWithUnvalidatedAccessTokenVisas() {
+    var validationCutoff =
+        new Timestamp(
+            Instant.now().minus(externalCredsConfig.getTokenValidationFrequency()).toEpochMilli());
```
dvoet: I would classify this as business logic and should be in manager code. Pass this cutoff value as a parameter to this function and have the config lookup and data math in the manager.

service/src/main/java/bio/terra/externalcreds/dataAccess/GA4GHPassportDAO.java
```diff
@@ -67,6 +76,25 @@ public GA4GHPassport insertPassport(GA4GHPassport passport) {
...
+    var namedParameters = new MapSqlParameterSource("validationCutoff", validationCutoff);
+
+    var query =
+        "SELECT DISTINCT la.id as linked_account_id, la.provider_id, p.jwt as passport_jwt FROM linked_account la"
+            + " JOIN ga4gh_passport p"
+            + " ON p.linked_account_id = la.id"
+            + " LEFT JOIN ga4gh_visa v"
```
dvoet: I don't think this needs to be `LEFT`

service/src/main/java/bio/terra/externalcreds/dataAccess/LinkedAccountDAO.java
```diff
@@ -89,6 +97,17 @@ public boolean deleteLinkedAccountIfExists(String userId, String providerId) {
     return jdbcTemplate.update(query, namedParameters) > 0;
   }
 
+  public boolean updateLinkAuthenticationStatus(int linkedAccountId, boolean isAuthenticated) {
```
dvoet: can this be removed?

service/src/main/java/bio/terra/externalcreds/services/ProviderService.java
```diff
@@ -183,40 +184,38 @@ public void deleteLink(String userId, String providerId) {
...
+
+          // If the response is not "valid", get a new passport.
+          if (!responseBody.equalsIgnoreCase("valid")) {
+            log.info(
+                String.format(
+                    "Found a visa that is not valid, with validation response '%s'. Refreshing passport.",
+                    responseBody));
+            var linkedAccount = linkedAccountService.getLinkedAccount(pd.getLinkedAccountId());
+            try {
+              authAndRefreshPassport(linkedAccount.get());
```
dvoet: calling `.get()` like this is typically a bad idea. There really should be a check to ensure the `Optional` isn't empty first. In this case, there is a window of time between the call to `passportService.getPassportsWithUnvalidatedAccessTokenVisas` and `linkedAccountService.getLinkedAccount` where the linked account may have been deleted. So if this gets an `Empty` the code should maybe log a message and continue on without error.

service/src/main/java/bio/terra/externalcreds/services/ProviderService.java
```diff
@@ -238,6 +236,59 @@ void authAndRefreshPassport(LinkedAccount linkedAccount) {
...
+  }
+
+  @VisibleForTesting
+  String validatePassportWithProvider(PassportVerificationDetails passportDetails) {
+    var validationEndpoint =
+        externalCredsConfig
+            .getProviders()
+            .get(passportDetails.getProviderId())
+            .getValidationEndpoint()
+            .get();
```
dvoet: what happens if this is empty?

service/src/test/java/bio/terra/externalcreds/dataAccess/LinkedAccountDAOTest.java
```diff
@@ -39,7 +49,7 @@ void testInsertAndGetLinkedAccount() {
 
     var loadedLinkedAccount =
         linkedAccountDAO.getLinkedAccount(linkedAccount.getUserId(), linkedAccount.getProviderId());
-    assertEquals(Optional.of(savedLinkedAccount), loadedLinkedAccount);
+    assertEquals(savedLinkedAccount, loadedLinkedAccount.get());
```
dvoet: this (and the next) was better as it was written before

dvoet (changes requested):

service/src/test/java/bio/terra/externalcreds/dataAccess/LinkedAccountDAOTest.java
Shakespeared: Cool, I'll put those back. I thought .get() might be neater but that makes sense!

service/src/main/java/bio/terra/externalcreds/dataAccess/GA4GHPassportDAO.java
Shakespeared: By "in manager code" do you mean the "back-ecm" code?

dvoet: sorry, I meant in _Service_ code, i.e. in `PassportService`

service/src/main/java/bio/terra/externalcreds/services/ProviderService.java
```diff
@@ -183,40 +184,42 @@ public void deleteLink(String userId, String providerId) {
...
+                String.format(
+                    "Found a visa that is not valid, with validation response '%s'. Refreshing passport.",
+                    responseBody));
+            var linkedAccount = linkedAccountService.getLinkedAccount(pd.getLinkedAccountId());
+            try {
+              if (linkedAccount.isEmpty()) {
+                log.info("No linked account found when trying to validate passport.");
+              } else {
+                authAndRefreshPassport(linkedAccount.get());
+              }
```
dvoet:
```suggestion
              linkedAccount.ifPresentOrElse(
                  this::authAndRefreshPassport,
                  () -> log.info("No linked account found when trying to validate passport."));
```

dvoet (changes requested):
one suggestion on the new code then 2 other changes:
1. make `validatePassportsWithAccessTokenVisas` return the size of `passportDetailsList`
2. add the call to `validatePassportsWithAccessTokenVisas` in `ExternalCredsCronApplication` and update the `Map` in log message to include the count

service/src/main/java/bio/terra/externalcreds/dataAccess/GA4GHPassportDAO.java
Shakespeared: oh ok, gotcha!

service/src/main/java/bio/terra/externalcreds/ExternalCredsCronApplication.java
```diff
@@ -39,12 +39,21 @@ public ExternalCredsCronApplication(ProviderService providerService) {
...
         Map.of("expiring_passport_count", expiringPassportCount));
   }
+
+  @Scheduled(
+      fixedRateString = "#{${externalcreds.validation-background-job-interval-mins} * 60 * 1000}")
+  public void validateVisas() {
+    log.info("beginning validateVisas");
+    var checkedPassportCount = providerService.validatePassportsWithAccessTokenVisas();
+    log.info("completed validateVisas", Map.of("checked_passport_count", checkedPassportCount));
+  }
```
dvoet: I don't think we want another scheduled function, just put it in `checkForExpiringCredentials`

service/src/main/java/bio/terra/externalcreds/config/ExternalCredsConfigInterface.java
```diff
@@ -20,10 +20,14 @@
...
+
+  Duration getVisaAndPassportRefreshDuration();
 
   /** List of URIs that are allowable in jku headers of JWTs */
   Collection<URI> getAllowedJwksUris();
 
-  int getBackgroundJobIntervalMins();
+  int getRefreshBackgroundJobIntervalMins();
+
+  int getValidationBackgroundJobIntervalMins();
```
dvoet, line 32: what's this?

service/src/main/java/bio/terra/externalcreds/ExternalCredsCronApplication.java
Shakespeared: Are we not running those at different intervals? Or is there a way to do that within the same function?

dvoet: my main concern is concurrency between the 2 jobs, I think it makes sense to first make sure we refresh anything that is going to expire soon then check of any visas that need to be validated

dvoet: I don't think the intervals for running the background job need to be different. Each task has a different threshold for things to refresh/validate and the background job interval is how often we check for things that hit the threshold.

Shakespeared: In that case should we remove the separate config variable for the frequency of the validation job, or just leave it for now but use the refresh config interval for both?

dvoet: remove it

dvoet (approved):

PR #69, opened 2021-09-29
Fix visa validation [POL-89]

snwessel:
We had been making a POST request with the entire passport in the request body, but realized this approach doesn't actually work with the RAS endpoint. Instead, we need to make a GET request with only the visa as a query parameter. 

This PR is mostly renaming things from "passport" to "visa", including:
- `validatePassportsWithAccessTokenVisas` -> `validateAccessTokenVisas`
- `getPassportsWithUnvalidatedAccessTokenVisas` -> `getUnvalidatedAccessTokenVisaDetails` (also moving this and its tests from the PassportDAO to the VisaDAO).
- `validatePassportWithProvider` -> `validateVisaWithProvider`

The only non-naming changes are:
- updating the database query to return the visa JWT instead of the passport JWT
- updating the structure of the structure of the request we make to the RAS validation endpoint

### DIFF
--- a/service/src/main/java/bio/terra/externalcreds/dataAccess/GA4GHVisaDAO.java
+++ b/service/src/main/java/bio/terra/externalcreds/dataAccess/GA4GHVisaDAO.java
@@ -67,16 +67,17 @@ public class GA4GHVisaDAO {
 
   public List<VisaVerificationDetails> getUnvalidatedAccessTokenVisaDetails(
       Timestamp validationCutoff) {
-    var namedParameters = new MapSqlParameterSource("validationCutoff", validationCutoff);
+    var namedParameters =
+        new MapSqlParameterSource("tokenType", TokenTypeEnum.access_token.toString())
+            .addValue("validationCutoff", validationCutoff);
 
     var query =
-        "SELECT DISTINCT la.id as linked_account_id, la.provider_id, v.jwt FROM linked_account la"
+        "SELECT DISTINCT la.id as linked_account_id, la.provider_id as provider_id, v.jwt as jwt FROM linked_account la"
             + " JOIN ga4gh_passport p"
             + " ON p.linked_account_id = la.id"
             + " JOIN ga4gh_visa v"
             + " ON v.passport_id = p.id"
-            + " WHERE v.token_type = "
-            + String.format("'%s'", TokenTypeEnum.access_token)
+            + " WHERE v.token_type = :tokenType::token_type_enum"
             + " AND v.last_validated <= :validationCutoff";
 
     return jdbcTemplate.query(query, namedParameters, new VisaVerificationDetailsRowMapper());
--- a/service/src/main/java/bio/terra/externalcreds/dataAccess/GA4GHVisaDAO.java
+++ b/service/src/main/java/bio/terra/externalcreds/dataAccess/GA4GHVisaDAO.java
@@ -2,8 +2,10 @@ package bio.terra.externalcreds.dataAccess;
 
 import bio.terra.externalcreds.models.GA4GHVisa;
 import bio.terra.externalcreds.models.TokenTypeEnum;
+import bio.terra.externalcreds.models.VisaVerificationDetails;
 import java.sql.ResultSet;
 import java.sql.SQLException;
+import java.sql.Timestamp;
 import java.sql.Types;
 import java.util.List;
 import java.util.Optional;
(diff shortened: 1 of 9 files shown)


service/src/main/java/bio/terra/externalcreds/dataAccess/GA4GHVisaDAO.java
```diff
@@ -63,6 +65,23 @@ public GA4GHVisa insertVisa(GA4GHVisa visa) {
...
+    var namedParameters = new MapSqlParameterSource("validationCutoff", validationCutoff);
+
+    var query =
+        "SELECT DISTINCT la.id as linked_account_id, la.provider_id, v.jwt FROM linked_account la"
+            + " JOIN ga4gh_passport p"
+            + " ON p.linked_account_id = la.id"
+            + " JOIN ga4gh_visa v"
+            + " ON v.passport_id = p.id"
+            + " WHERE v.token_type = "
+            + String.format("'%s'", TokenTypeEnum.access_token)
```
dvoet: I know this was copied, but can you change this to a named parameter? String concatenation in sql is frowned upon because it leads to sql injection attacks even though I know that is not a problem in this case.

service/src/main/java/bio/terra/externalcreds/dataAccess/GA4GHVisaDAO.java
```diff
@@ -63,6 +65,23 @@ public GA4GHVisa insertVisa(GA4GHVisa visa) {
     return jdbcTemplate.query(query, namedParameters, new GA4GHVisaRowMapper());
   }
 
+  public List<VisaVerificationDetails> getUnvalidatedAccessTokenVisaDetails(
+      Timestamp validationCutoff) {
+    var namedParameters = new MapSqlParameterSource("validationCutoff", validationCutoff);
+
+    var query =
+        "SELECT DISTINCT la.id as linked_account_id, la.provider_id, v.jwt FROM linked_account la"
```
dvoet: I would include the aliases for provider_id and jwt even though they are the defaults because they are important for the row mapper

service/src/main/java/bio/terra/externalcreds/services/ProviderService.java
```diff
@@ -184,12 +184,12 @@ public void deleteLink(String userId, String providerId) {
     linkedAccountService.deleteLinkedAccount(userId, providerId);
   }
 
-  public int validatePassportsWithAccessTokenVisas() {
-    var passportDetailsList = passportService.getPassportsWithUnvalidatedAccessTokenVisas();
+  public int validateAccessTokenVisas() {
+    var visaDetailsList = passportService.getUnvalidatedAccessTokenVisaDetails();
 
-    passportDetailsList.forEach(
+    visaDetailsList.forEach(
```
dvoet: I would actually change this implementation a bit. The problem is that a passport may have multiple visas and if more than one is invalid we refresh the entire passport more than once. So I would change this to 2 steps. First find all the linked accounts that need refreshing:
```
    var linkedAccountIdsToRefresh = visaDetailsList.stream().flatMap(visaDetails -> {
      var responseBody = validateVisaWithProvider(visaDetails);
      return responseBody.equalsIgnoreCase("valid") ? Stream.empty() : Stream.of(visaDetails.getLinkedAccountId());
    }).distinct();
```
Then for each of those do the block of code starting at line 196 and log each linked account id being refreshed

service/src/main/java/bio/terra/externalcreds/services/ProviderService.java
```diff
@@ -270,12 +270,11 @@ public LinkedAccountWithPassportAndVisas getRefreshedPassportsAndVisas(
   }
 
   @VisibleForTesting
-  String validatePassportWithProvider(PassportVerificationDetails passportDetails) {
-    var providerProperties =
-        externalCredsConfig.getProviders().get(passportDetails.getProviderId());
+  String validateVisaWithProvider(VisaVerificationDetails visaDetails) {
```
dvoet, line 276: I would log the response (no matter what it is) at the end of this function including the linked account id and provider id and forego the log statement on 197

dvoet (changes requested):
for log messages consider using structure logging instead of embedding details in the message string:
`log.info("some message", Map.of(key, value, …)`
These won't show up in the human readable log messages but will in json log messages that go to google logging and are much more searchable

service/src/main/java/bio/terra/externalcreds/services/ProviderService.java
snwessel: @dvoet is this change what you had in mind? https://github.com/DataBiosphere/terra-external-credentials-manager/pull/69/commits/c9f33c4ff978db47eceff91a18865f1414c13c72

service/src/main/java/bio/terra/externalcreds/services/ProviderService.java
```diff
@@ -184,30 +186,39 @@ public void deleteLink(String userId, String providerId) {
...
+        visaDetailsList.stream()
+            .flatMap(
+                visaDetails -> {
+                  var responseBody = validateVisaWithProvider(visaDetails);
+                  log.info(
+                      "Got visa validation response: {}",
+                      Map.of(
+                          "linkedAccountId", visaDetails.getLinkedAccountId(),
+                          "providerId", visaDetails.getProviderId(),
+                          "validationResponse", responseBody));
```
dvoet: Yes, this is the log message I had in mind but you should not have the `: {}` at the end. Also, I would put this log statement in `validateVisaWithProvider`, then you can inline `responseBody` and this block of code gets much cleaner.

dvoet (approved):
a little jig then 👍

Shakespeared (approved):

PR #72, opened 2021-09-30
rename providerId/provider to providerName

Shakespeared:
Touches a lot of files but this is a straight rename, should not be any functional code changes. I thiiiiink I got everything, all instances should be consistent now. Also renamed the db column in from provider_id to provider_name.

### DIFF
--- a/service/src/main/resources/db/changelog/changesets/20210930_rename_provider_id.yaml
+++ b/service/src/main/resources/db/changelog/changesets/20210930_rename_provider_id.yaml
@@ -4,5 +4,4 @@ databaseChangeLog:
       author: kfeldman
       changes:
         - sql:
-            sql: >
-              ALTER TABLE linked_account RENAME COLUMN provider_id TO provider_name;
+            sql: ALTER TABLE linked_account RENAME COLUMN provider_id TO provider_name;
--- a/service/src/main/resources/db/changelog/changesets/20210930_rename_provider_id.yaml
+++ b/service/src/main/resources/db/changelog/changesets/20210930_rename_provider_id.yaml
@@ -4,4 +4,5 @@ databaseChangeLog:
       author: kfeldman
       changes:
         - sql:
-            sql: ALTER TABLE linked_account RENAME COLUMN provider_id TO provider_name;
+            sql: >
+              ALTER TABLE linked_account RENAME COLUMN provider_id TO provider_name;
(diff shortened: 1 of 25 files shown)


service/src/main/resources/db/changelog/changesets/20210930_rename_provider_id.yaml
```diff
@@ -0,0 +1,8 @@
+databaseChangeLog:
+  - changeSet:
+      id: "rename_provider_id_column"
+      author: kfeldman
+      changes:
+        - sql:
+            sql: >
+              ALTER TABLE linked_account RENAME COLUMN provider_id TO provider_name;
```
zarsky-broad, line 8: Should this just be one line?
```suggestion
            sql: ALTER TABLE linked_account RENAME COLUMN provider_id TO provider_name;
```

service/src/main/resources/db/changelog/db.changelog-master.yaml
```diff
@@ -14,3 +14,7 @@ databaseChangeLog:
   - include:
       file: changesets/20210820_add_is_authenticated_column.yaml
       relativeToChangelogFile: true
+  - include:
+      file: changesets/20210930_rename_provider_id.yaml
+      relativeToChangelogFile: true
+
```
zarsky-broad:
```suggestion
```

zarsky-broad (approved):
Looks solid, and the fact that the tests all passed would indicate that if you missed anywhere, you at least missed it consistently

service/src/main/resources/db/changelog/changesets/20210930_rename_provider_id.yaml
Shakespeared: yep probably

Shakespeared: actually it looks like our other one liners aren't inlined, should we be consistent?

zarsky-broad: After poking around, we have a frustrating mix of `>`, `"`, and neither delimiting our changeset sql strings. Your call.

Shakespeared: Cool cool, maybe I'll just make 'em consistent then. My inclination is to use `>` because most of the multi-line ones are already handled that way

snwessel (approved):

PR #73, opened 2021-10-01
fix bug with duplicate authorizations and added role check

dvoet:


service/src/main/java/bio/terra/externalcreds/visaComparators/RASv1_1.java
```diff
@@ -47,7 +48,8 @@ public boolean authorizationsMatch(GA4GHVisa visa1, GA4GHVisa visa2) {
       throws ParseException, JsonProcessingException {
     var visaClaim =
         JWTParser.parse(visa.getJwt()).getJWTClaimsSet().getClaim(DBGAP_CLAIM).toString();
-    var visaPermissions = Set.of(objectMapper.readValue(visaClaim, DbGapPermission[].class));
```
dvoet, line 50: it turns out `Set.of()` throws an exception when there are dups :(

service/src/main/java/bio/terra/externalcreds/visaComparators/RASv1_1.java
```diff
@@ -69,6 +71,8 @@ public boolean visaTypeSupported(GA4GHVisa visa) {
     @JsonProperty("consent_group")
     String getConsentGroup();
 
+    String getRole();
```
dvoet, line 74: I added role as a required field to detect when it changes. I don't think it really matters but seems better to send a notification and let downstream systems figure it out.

Shakespeared (approved):

snwessel (approved):
LGTM!

PR #76, opened 2021-10-05
choose any free port

dvoet:
I suspect an intermittent failure stems from reusing the same port in each test for a mock server. This change lets the mock server choose any free port.

suggest ignoring whitespace

### DIFF
--- a/service/src/test/java/bio/terra/externalcreds/services/ProviderServiceTest.java
+++ b/service/src/test/java/bio/terra/externalcreds/services/ProviderServiceTest.java
@@ -101,34 +101,32 @@ public class ProviderServiceTest extends BaseTest {
     }
 
     private void testWithRevokeResponseCode(HttpStatus httpStatus) {
-      var revocationPath = "/test/revoke/";
-      var mockServerPort = 50555;
-      var linkedAccount = TestUtils.createRandomLinkedAccount();
+      try (var mockServer = ClientAndServer.startClientAndServer()) {
+        var revocationPath = "/test/revoke/";
+        var linkedAccount = TestUtils.createRandomLinkedAccount();
 
-      var providerInfo =
-          TestUtils.createRandomProvider()
-              .setRevokeEndpoint(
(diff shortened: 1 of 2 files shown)


service/src/test/java/bio/terra/externalcreds/services/ProviderServiceTest.java
```diff
@@ -139,8 +137,6 @@ private void testWithRevokeResponseCode(HttpStatus httpStatus) {
         providerService.deleteLink(linkedAccount.getUserId(), linkedAccount.getProviderName());
         verify(linkedAccountServiceMock)
             .deleteLinkedAccount(linkedAccount.getUserId(), linkedAccount.getProviderName());
-      } finally {
-        mockServer.stop();
```
snwessel, line 143: The `JwtSigningTestUtils` has a teardown method that stops the mockServer: 
```
public void tearDown() { mockServer.stop(); }
```
Should this test class have something similar since you're deleting `mockServer.stop()` line here?

dvoet: the difference is that in `JwtSigningTestUtils` all the tests use the same mock server

snwessel: Okay that makes sense. In that case, should we keep the `mockServer.stop()` line above ^ in `ProviderServiceTest`?

Shakespeared (approved):
LGTM

service/src/test/java/bio/terra/externalcreds/services/ProviderServiceTest.java
dvoet: Notice I also changed the try on line 104. The mock server is auto closable and will be closed automatically at the end of the try.

snwessel: Thanks for explaining! That makes sense.

snwessel (approved):

PR #78, opened 2021-10-05
Mock test configs [POL-91]

snwessel:
I've been looking into test config mocking and came to a few conclusions:
1. It's not worth creating test utils for this because the config mocking is pretty simple to begin with. 
2. We often forget to mock the non-provider configs, so our unit tests often use the actual config values. To me, that seems like a weird dependency which could cause our tests could break when the config values are updated. 

The changes below create a separate set of configs which can be used by the tests. That way we don't need to remember to mock them every time they're relevant to a test.

### DIFF
--- a/service/src/test/java/bio/terra/externalcreds/services/ProviderServiceTest.java
+++ b/service/src/test/java/bio/terra/externalcreds/services/ProviderServiceTest.java
@@ -403,10 +403,6 @@ public class ProviderServiceTest extends BaseTest {
       passportDAO.insertPassport(expiringPassport);
       passportDAO.insertPassport(notExpiringPassport);
 
-      // mock the configs
-      when(externalCredsConfigMock.getVisaAndPassportRefreshDuration())
-          .thenReturn(Duration.ofMinutes(30));
-
       // check that authAndRefreshPassport is called exactly once with the expiring linked account
       var providerServiceSpy = Mockito.spy(providerService);
       providerServiceSpy.refreshExpiringPassports();
--- a/service/src/test/resources/application.yml
+++ b/service/src/test/resources/application.yml
@@ -1,5 +1,5 @@
+# Override the ECM config values for tests so unit tests won't be impacted by config changes
 externalcreds:
-  # 
   sam-base-path: http://localhost:1000/
   background-job-interval-mins: 5
   token-validation-duration: 1h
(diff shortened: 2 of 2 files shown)


service/src/test/resources/application.yml
```diff
@@ -0,0 +1,17 @@
...
+  background-job-interval-mins: 5
+  token-validation-duration: 1h
+  visa-and-passport-refresh-duration: 30m
+  authorization-change-events-enabled: false
+
+spring:
+  datasource:
+    password: ${DATABASE_USER_PASSWORD:ecmpwd}
+    url: jdbc:postgresql://${DATABASE_HOSTNAME:localhost}:5432/${DATABASE_NAME:ecm}
+    username: ${DATABASE_USER:ecmuser}
```
snwessel, line 13: I haven't found a way to avoid including the `spring.datasource` definition here. If anyone has an idea of how to simplify this, let me know!

Shakespeared (approved):

service/src/test/java/bio/terra/externalcreds/services/ProviderServiceTest.java
```diff
@@ -403,10 +403,6 @@ void testOnlyExpiringPassportsAreRefreshed() {
       passportDAO.insertPassport(expiringPassport);
       passportDAO.insertPassport(notExpiringPassport);
 
-      // mock the configs
-      when(externalCredsConfigMock.getVisaAndPassportRefreshDuration())
-          .thenReturn(Duration.ofMinutes(30));
-
```
dvoet, line 409: I am concerned about removing this because the test cares about this value. Moving the value to application.yml creates a dependency that is hard to see.

dvoet (changes requested):
I agree that the dependency between the production config and tests seems weird. But I think most tests don't actually care what the config is, i.e. they work for all config values. I think the tests that are sensitive to particular config values should have those values baked into the test itself to eliminate dependencies.

The benefit to loading production config in tests is that something in our test path actually does so and we know the config is valid. We could have a separate test just for that. If we add a required config value it would be great to catch in unit tests that a value is missing from prod config.

service/src/test/java/bio/terra/externalcreds/services/ProviderServiceTest.java
snwessel: That makes sense, and I totally agree that it does create a dependency that's hard to see. I think that there are some issues with the way we're currently mocking (like how it's easy to forget to mock something and create a weird dependency), but I'm also not sure of a better solution for this right now, so I might just close this ticket and spend some more time thinking about it.

PR #88, opened 2022-01-04
POL-151 don't return expired passports

dvoet:
https://broadworkbench.atlassian.net/browse/POL-151

This check could have been done in the DAO (i.e. sql where clause) but the DAO should be general purpose, without business logic.

This check could also have been done in the `PassportService` but that seemed like it would be unexpected. I can certainly envision other code in ECM that needs to get the passport, expired or not, that would be surprised by this condition.

So controller logic seemed most appropriate.


snwessel (approved):
Makes sense to me!

PR #91, opened 2022-01-06
[POL-98] Part 1: Perf Test Suite Setup

snwessel:
Setting up a performance test suite that runs locally, which includes the following changes:
- Adding a `ListProviders` test. This is the only other endpoint that's straightforward to test so I figured it might as well be included.
- Adding a config for each individual perf test. These are identical to the integration test configs, except that they run the test 10 times instead of once (yes it's a lot of copy and pasting, but that seems to be what test runner requires).
- Adding a `FullPerf` test suite, which just lists all of the configs we consider part of the full perf test

### DIFF
--- a/integration/src/main/java/scripts/testscripts/ListProviders.java
+++ b/integration/src/main/java/scripts/testscripts/ListProviders.java
@@ -2,7 +2,6 @@ package scripts.testscripts;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertNotNull;
-import static org.junit.jupiter.api.Assertions.assertTrue;
 
 import bio.terra.externalcreds.api.OidcApi;
 import bio.terra.testrunner.runner.TestScript;
@@ -29,6 +28,5 @@ public class ListProviders extends TestScript {
 
     // check the response body
     assertNotNull(providers);
-    assertTrue(providers.size() > 0);
   }
 }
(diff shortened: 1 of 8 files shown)


integration/src/main/java/scripts/testscripts/ListProviders.java
```diff
@@ -0,0 +1,34 @@
...
+import scripts.utils.ClientTestUtils;
+
+@Slf4j
+public class ListProviders extends TestScript {
+
+  @Override
+  public void userJourney(TestUserSpecification testUser) throws Exception {
+    log.info("Checking the version endpoint.");
+    var apiClient = ClientTestUtils.getClientWithoutAccessToken(server);
+    var oidcApi = new OidcApi(apiClient);
```
snwessel, line 20: The `OidcApi` usually requires authentication through the proxy but it seems the proxy isn't used when running integration or perf tests locally - so we get by here without authenticating.

Shakespeared (approved):
LGTM, although I don't know about the proxy question

integration/src/main/java/scripts/testscripts/ListProviders.java
snwessel: After sleeping on this, I think it makes sense that there's no proxy to require auth locally. The next step will be to run these tests against perf - and I'm pretty sure that the proxy _will_ require authentication in that case, so I'll probably need to add an access token once we set that up (in my next PR if that's okay with reviewers 😄).

zarsky-broad (approved):

PR #98, opened 2022-01-27
add perf test for getProviderPassport [POL-100]

Shakespeared:
This adds a perf test for the getProviderPassport endpoint. I left a few TODOs to indicate things we may need to adjust as we expand our perf testing. 

The exact config we want for this test may change, especially as there are improvements to the TestRunner configs in progress, so this PR is primarily for the test itself. Includes a few minor changes to the README as well.


integration/src/main/resources/configs/perf/GetProviderPassport.json
```diff
@@ -0,0 +1,17 @@
+{
+  "name": "GetProviderPassport",
+  "description": "Gets a user's passport from the specified provider.",
+  "serverSpecificationFile": "ecm-local.json",
+  "kubernetes": {},
+  "application": {},
+  "testScripts": [
+    {
+      "name": "GetProviderPassport",
+      "numberOfUserJourneyThreadsToRun": 1000,
```
Shakespeared, line 10: NB: This config gets close to what we want, but will probably need to be updated in the relatively near future as there is work upcoming from the PF team on better ramp-up load configs for TestRunner

dvoet (approved):

integration/src/main/java/scripts/testscripts/GetProviderPassport.java
```diff
@@ -0,0 +1,43 @@
...
+import scripts.utils.ClientTestUtils;
+
+@Slf4j
+public class GetProviderPassport extends TestScript {
+  private String provider;
+  private OidcApi oidcApi;
+
+  @Override
+  public void setup(List<TestUserSpecification> testUsers) throws Exception {
+    // TODO: do we want to loop through this list in case we want multiple test users in the future?
```
snwessel, line 21: I was wondering about this too. @dvoet do you have any preference on whether we use more than one test user to authenticate for these tests? I am currently only adding the one actual Sam account to our generated data but I could add in more if we wanted.

integration/src/main/java/scripts/testscripts/GetProviderPassport.java
```diff
@@ -0,0 +1,43 @@
...
+    // TODO: update GetProviderPassport.json in perf to run 120 tests per second
+    // check the response code
+    var httpCode = oidcApi.getApiClient().getStatusCode();
+    var passport = oidcApi.getProviderPassport(provider);
+
+    assertEquals(HttpStatusCodes.STATUS_CODE_OK, httpCode);
+    log.info("Get provider passport return code: {}", httpCode);
+
+    // check the response body
+    assertNotNull(provider);
```
snwessel, line 41: I know this is a bit of a nit pick but should we check that the provider's not null before using it to make the request? Right now, if it were ever null it would lead to a bit of a confusing error message (that the passport endpoint returned an unhappy status code).

Shakespeared: Added a null check!

snwessel (approved):
I'll make a note to have a follow up conversation about adding more Sam users (and we can make a separate ticket for that if we decide to do it)

PR #99, opened 2022-01-28
[POL-156] Generate data for performance tests

snwessel:
This script does the following:
- Deletes all of the existing rows from the database
- Inserts 10,000 linked accounts and passports, each with a random user id
- Inserts a linked account and passport for the `Scarlett.Flowerpicker@test.firecloud.org` testUser (which is already set up in Sam).

It can be run on the perf db by using the existing psql-connect script ([link here](https://github.com/broadinstitute/dsp-scripts/blob/master/firecloud/psql-connect.sh)).

### DIFF
--- a/integration/src/main/java/scripts/datageneration/regenerate_perf_data.sql
+++ b/integration/src/main/java/scripts/datageneration/regenerate_perf_data.sql
@@ -1,16 +1,35 @@
+-- This function inserts a semi-random linked account and passport with
+-- the given user_id and external_user_id
+CREATE PROCEDURE insert_test_record(user_id text, external_user_id text)  
+LANGUAGE SQL 
+AS $$ 
+    -- Insert a linked account with the given values
+    INSERT INTO linked_account 
+    (id, user_id, provider_name, refresh_token, expires, external_user_id, is_authenticated)
+    VALUES (
+        nextval('linked_account_id_seq'), 
+        user_id, 
+        'ras', 
+        'testToken', 
+        current_timestamp + interval '3000 year', 
+        external_user_id, 
+        true
+    );
+    -- Insert a corresponding passport
+        INSERT INTO ga4gh_passport (linked_account_id, jwt, expires)
+        VALUES (
+            currval('linked_account_id_seq'), 
+            'testJwt', 
+            current_timestamp + interval '3000 year'
+        );
+$$;  
+
 
 -- delete all of the existing data
 TRUNCATE linked_account, ga4gh_passport, ga4gh_visa;
 
 -- Insert Linked Accounts and Passports for specific users that already exist in SAM
-INSERT INTO linked_account 
-(id, user_id, provider_name, refresh_token, expires, external_user_id, is_authenticated)
-VALUES
-(nextval(pg_get_serial_sequence('linked_account','id')), '114925642006220098835', 'ras', 'testToken', current_timestamp + interval '3000 year', 'Scarlett.Flowerpicker@test.firecloud.org', true);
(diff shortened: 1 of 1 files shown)


integration/src/main/java/scripts/datageneration/regenerate_perf_data.sql
```diff
@@ -0,0 +1,47 @@
+-- This function inserts a semi-random linked account and passport with
+-- the given user_id and external_user_id
+CREATE PROCEDURE insert_test_record(user_id text, external_user_id text)
+LANGUAGE SQL
+AS $$
```
Shakespeared, line 12: Can you explain the usage of the $$s? I haven't seen the double bling before (and it's hard to google)

integration/src/main/java/scripts/datageneration/regenerate_perf_data.sql
```diff
@@ -0,0 +1,47 @@
...
+    'testToken',
+    current_timestamp + interval '3000 year',
+    external_user_id,
+    true
+  );
+  -- Insert a corresponding passport
+  INSERT INTO ga4gh_passport (linked_account_id, jwt, expires)
+  VALUES (
+    currval('linked_account_id_seq'),
+    'testJwt',
```
dvoet: make this a more realistic jwt. I am especially concerned that it is a realistic size.

dvoet (changes requested):
also please insert visas

integration/src/main/java/scripts/datageneration/regenerate_perf_data.sql
snwessel: Here's some documentation on it: https://www.geeksforgeeks.org/postgresql-dollar-quoted-string-constants/

I think it's only really used for functions and stored procedures though (to enclose blocks of code that will be reused).

integration/src/main/java/scripts/datageneration/regenerate_perf_data.sql
snwessel: I looked into a few different options for doing this but postgresql is pretty limited in what it can actually do:
1. You had suggested reading a static value from a file. Unfortunately, there doesn't seem to be a way to read from a file path relative to the script location. Instead, it reads relative to the location of the postgres data directory, which isn't helpful. 
2. If we want an actual JWT, we could just hard code the JWT string in the SQL script. This is definitely the simplest solution, even though I do prefer to avoid hard-coding things. 
3. If we only care about the string being a larger size and want to avoid hard coding, we could randomly generate a long string by doing something like [this](https://www.simononsoftware.com/random-string-in-postgresql/).

I would love to hear your thoughts on what you think is best!

snwessel: I ended up coming up with something that I think works pretty well to randomly generate a long string. Let me know what you think!

integration/src/main/java/scripts/datageneration/regenerate_perf_data.sql
```diff
@@ -0,0 +1,69 @@
+-- TODO: test out and debug again
```
dvoet: is this still a todo?

dvoet (approved):

integration/src/main/java/scripts/datageneration/regenerate_perf_data.sql
Shakespeared: TODO :)

Shakespeared (approved):
LGTM

PR #100, opened 2022-01-28
[POL-172] Integrate AppSec Trivy GitHub Action

dinvlad:
This PR adds [DSP AppSec Trivy GitHub Action](https://github.com/broadinstitute/dsp-appsec-trivy-action) for the Docker image.

The action will run on every PR and give you feedback if the Docker image contains critical OS vulnerabilities.

This is packaged as a standalone workflow so it doesn't break your other workflows, but if you'd like to integrate it into another workflow instead, please let us know.


.github/workflows/trivy.yaml
```diff
@@ -0,0 +1,35 @@
...
+        with:
+          path: |
+            ~/.gradle/caches
+            ~/.gradle/wrapper
+          key: v1-${{ runner.os }}-gradle-${{ github.ref }}-${{ github.sha }}
+          restore-keys: v1-${{ runner.os }}-gradle-${{ github.ref }}
+
+      - name: Build image locally with Jib
+        run: |
+          ./gradlew :service:jibDockerBuild --image="${image}" -Djib.console=plain
```
zarsky-broad, line 31: I'd expect this to fail, because it doesn't have the base image to build from. See here: https://github.com/DataBiosphere/terra-external-credentials-manager/blob/cc4dcee38c83319c7d1817a4e0b6d8fe6305c6c4/.github/workflows/publish.yml#L51-L65

dinvlad: Could you clarify? The builds appear to succeed..

dinvlad: Also if you already have Trivy action in that workflow, then nothing to do for us here, we can close it =)

zarsky-broad: Ah, it must just be building from the default base image, so it might still complete, but it wouldn't be scanning the actual produced image we'd be using.

If just running before an image is pushed is good enough, though, then we're already doing it and you can close this PR if you want.

dinvlad: OK cool, yes that was the intent of this PR as well (running before pushing is exactly what we want).

PR #104, opened 2022-03-03
Pol 174 validate passport

dvoet:
Checkout my first pass of what an openapi spec would look like for validate passports. Passports is an array because there may be more than one (although we will only have one for now). Criteria is an array because if there is more than one passport different criteria could be accepted.

I expect different types of criteria in the future which is what the `oneOf` thing is doing. The swagger codegen is a little wonky here. It automatically puts in a `type` property which is unlisted in the openapi itself. So I added a property called `type` so at least it shows up. If you run the included test what gets printed by the controller is
```
class ValidatePassportRequest {
    passports: [I am a passport]
    criteria: [class RASv11 {
        class VisaCriteria {
            type: null
            issuer: visa issuer
        }
        phsId: phs001234
        consentCode: c1
    }]
}
```
Notice `type` is null (because that value was consumed already in figuring out which class to instantiate). The generated java client will actually put the `type` field in there twice. But it all seems to work. The main ugliness is that within ECM we need to know that `type` will be null and don't use it.

WDYT?


service/src/main/java/bio/terra/externalcreds/visaComparators/RASv1_1.java
```diff
@@ -44,25 +45,46 @@ public boolean authorizationsMatch(GA4GHVisa visa1, GA4GHVisa visa2) {
     }
   }
 
+  @Override
+  public boolean matchesCriterion(GA4GHVisa visa, VisaCriterion criterion) {
+    try {
+      assert criterion instanceof RASv1Dot1Criterion;
```
Shakespeared: Is there a reason why it's sometimes named as `RASv1Dot1` and sometimes `RASv1_1`?

Shakespeared (approved):
I like it, overall makes sense to me!

service/src/main/java/bio/terra/externalcreds/visaComparators/RASv1_1.java
snwessel: I am also curious about that.

service/src/main/java/bio/terra/externalcreds/services/PassportService.java
```diff
@@ -46,4 +68,94 @@ public PassportService(
...
+                        .getLinkedAccountByPassportJwtId(p.getPassport().getJwtId())
+                        .stream())
+            .collect(Collectors.toSet());
+
+    if (linkedAccounts.isEmpty()) {
+      throw new BadRequestException("unknown user");
+    }
+    if (linkedAccounts.size() > 1) {
+      throw new BadRequestException(
+          "a single request validate passport can contain only passports from the same linked account");
```
snwessel:
```suggestion
          "a single request to validate passport can contain only passports from the same linked account");
```

snwessel (approved):
Looks good to me! Just found one possible typo

dvoet:
@broad-dsp-appsec please review new api to validate passports

TomConner:
Looks really solid -- I see you're validating and allow-listing mostly all the things. Two questions: 

1. Is there anything to validate `alg`?
2. Do we expect `aud` (in which case we should validate)?

dvoet:
> 1. Is there anything to validate `alg`?
> 2. Do we expect `aud` (in which case we should validate)?

I am not aware of any requirements for these

TomConner:
I don't think NIH uses `aud` for the moment.

For `alg` I'm just thinking about attacks and best practices in https://www.rfc-editor.org/info/rfc8725 - the attacks are presumably very difficult given all the validation of issuer and allow-listing where you get passports in the first place. But tampering with `alg` is a well-documented attack vector and if there's a configuration or parser flag that locks that down, it would be nice to add.

TomConner (approved):

PR #112, opened 2022-03-22
PF-1354: Add a post method to generate the ssh key

yuhuyoyo:
This change limits the risk as we prevent the user from storing their important existing ssh key in the database. Instead, this is just a terra key generated for the user.


dvoet (changes requested):
I don't think this api makes sense as a PUT anymore. Also all apis must be application/json

yuhuyoyo:
> don't think this api makes sense as a PUT anymore. Also all apis must be application/json

I did some search on PUT and POST. The method will create or update the key so I thought PUT makes sense. 
https://www.rfc-editor.org/rfc/rfc7231#section-4.3.4

dvoet:
First line of your link: 
  The PUT method requests that the state of the target resource be
   created or replaced with the state defined by the representation
   enclosed in the request message payload

The payload no longer represents the state of the ssh key pair.

I am actually not a fan of this approach. Why is ECM responsible for generating this key?

yuhuyoyo:
> I am actually not a fan of this approach. Why is ECM responsible for generating this key?

If the user set up their notebook with the terra key and the terra key is compromised, they can revoke just the terra key without affecting their ssh set up elsewhere (e.g. their local machine). It's also preferred on the UI to just generate a terra key for the user instead of uploading their personal ssh private key. 

@gjuggler

yuhuyoyo:
@dvoet please see https://docs.google.com/document/d/1OkSos8KMihRyBoyUUcgQKP6HAeRsmXqnELrTnNftUaE/edit#bookmark=id.ghhcyy1ejwwm this section and the roadmap section below as the conclusion of our verily internal discussion around the security concern.

dvoet:
My concern with this approach is that ECM's responsibility is storage and retrieval of credentials, not generating credentials. I liked the API the way it was before this PR. I am not opposed to generating keys for the user but is that the responsibility of ECM? If we want it as a convenience in ECM then add the POST call (which is what I thought you proposed) but leave the PUT as is.

cbookg:
> My concern with this approach is that ECM's responsibility is storage and retrieval of credentials, not generating credentials. I liked the API the way it was before this PR. I am not opposed to generating keys for the user but is that the responsibility of ECM? If we want it as a convenience in ECM then add the POST call (which is what I thought you proposed) but leave the PUT as is.

+1 to keeping the existing set/get and adding the generate as a convenience.

yuhuyoyo:
alright, it is added back

service/src/main/java/bio/terra/externalcreds/services/SshKeyPairService.java
```diff
@@ -37,7 +56,46 @@ public SshKeyPairInternal putSshKeyPair(
...
+    dos.writeInt("ssh-rsa".getBytes(StandardCharsets.UTF_8).length);
+    dos.write("ssh-rsa".getBytes(StandardCharsets.UTF_8));
+    dos.writeInt(rsaPublicKey.getPublicExponent().toByteArray().length);
+    dos.write(rsaPublicKey.getPublicExponent().toByteArray());
+    dos.writeInt(rsaPublicKey.getModulus().toByteArray().length);
+    dos.write(rsaPublicKey.getModulus().toByteArray());
+    var publicKeyEncoded =
+        new String(Base64.encodeBase64(byteOs.toByteArray()), StandardCharsets.UTF_8);
+    return DEFAULT_PUBLIC_KEY_BEGIN + " " + publicKeyEncoded + " " + user;
+  }
```
dvoet: A few comments for this function
1. use `var`
2. do you need to close the output stream?
3. should the constant be used for all instances of `"ssh-rsa"`?
4. can you document what this is doing or provide a link?

service/src/main/java/bio/terra/externalcreds/services/SshKeyPairService.java
```diff
@@ -37,7 +56,46 @@ public SshKeyPairInternal putSshKeyPair(
...
+          new SshKeyPairInternal.Builder()
+              .privateKey(encodeRSAPrivateKey((RSAPrivateKey) rsaKeyPair.getPrivate()))
+              .publicKey(
+                  encodeRSAPublicKey((RSAPublicKey) rsaKeyPair.getPublic(), externalUserEmail))
+              .externalUserEmail(externalUserEmail)
+              .type(type)
+              .userId(userId)
+              .build());
+    } catch (NoSuchAlgorithmException | IOException e) {
+      throw new ExternalCredsException(e, HttpStatus.INTERNAL_SERVER_ERROR);
```
dvoet: `INTERNAL_SERVER_ERROR` is the default, no need to specify here

service/src/test/java/bio/terra/externalcreds/controllers/SshKeyApiControllerTest.java
```diff
@@ -108,18 +145,11 @@ void samThrowsApiException() throws Exception {
...
-    var rsaEncodedKeyPair = TestUtils.getRSAEncodedKeyPair(externalUserEmail);
-    var sshKeyPair =
-        new SshKeyPair()
-            .privateKey(rsaEncodedKeyPair.getLeft())
-            .publicKey(rsaEncodedKeyPair.getRight())
-            .externalUserEmail(externalUserEmail);
-    var requestBody = objectMapper.writeValueAsString(sshKeyPair);
     mvc.perform(
-            put("/api/sshkeypair/v1/{type}", sshKeyPairType)
+            post("/api/sshkeypair/v1/{type}", sshKeyPairType)
```
cbookg, line 149: why move these tests from the 'set' to the 'generate'?  do we want them on both, or just leave this on 'set'?

cbookg (approved):

service/src/test/java/bio/terra/externalcreds/controllers/SshKeyApiControllerTest.java
yuhuyoyo: shouldn't really matter which method we are calling because it's the same endpoint. I actually like using the post one then put as there's less set up.

dvoet (changes requested):
you should also get approval from appsec since this has a high security impact

yuhuyoyo:
> you should also get approval from appsec since this has a high security impact

I don't see how this have more security impact than what's already implemented (PUT). Also i have audit logging and kms encryption lining up after this PR.

dinvlad:
These are good points - let's discuss first in our follow-up meeting next week, as already mentioned!

yuhuyoyo:
> These are good points - let's discuss first in our follow-up meeting next week, as already mentioned!

thanks for commenting here despite you are ooo. I can set up a meeting to talk about KMS encryption. this PR is not related to KMS though so it shouldn't be blocked?

dinvlad:
> thanks for commenting here despite you are ooo. I can set up a meeting to talk about KMS encryption. this PR is not related to KMS though so it shouldn't be blocked?

I think it actually is - Doug's point that ECM should not be in charge of generating SSH keys aligns well with what I'm asking to add for client-side encryption. But happy to discuss this next week - feel free to send an invite!

dvoet (approved):

yuhuyoyo:
I have scheduled follow-on description next week for data encryption. For verily environment, there will be no deployment beyond devel in the meantime.

PR #114, opened 2022-03-28
POL-178 validate redirect uris

dvoet:


service/src/main/java/bio/terra/externalcreds/services/ProviderService.java
```diff
@@ -140,6 +142,15 @@ public int refreshExpiringPassports() {
             });
   }
 
+  private void validateRedirectUri(String redirectUri, ProviderProperties providerInfo) {
```
Shakespeared, line 145: What's the reason for filtering this list as opposed to looking for a match?

Shakespeared (approved):
LGTM just left a question

service/src/main/java/bio/terra/externalcreds/services/ProviderService.java
dvoet: I was being dense. I found `noneMatch`!

zarsky-broad:
Any reason for not using `var` with `ProviderProperties`?

zarsky-broad (approved):
Code looks good, tests seem comprehensive.

sarahgibs (approved):
Looks good.

PR #116, opened 2022-03-29
java 17 [POL-217]

zarsky-broad:


dvoet (approved):

service/src/main/java/bio/terra/externalcreds/controllers/OidcApiController.java
```diff
@@ -25,32 +25,15 @@
...
-      AuditLogger auditLogger) {
-    this.request = request;
-    this.linkedAccountService = linkedAccountService;
-    this.mapper = mapper;
-    this.passportService = passportService;
-    this.providerService = providerService;
-    this.samService = samService;
-    this.auditLogger = auditLogger;
-  }
+public record OidcApiController(
```
dvoet, line 28: nifty

dvoet (approved):

service/src/main/java/bio/terra/externalcreds/controllers/OidcApiController.java
zarsky-broad: Of note: records are implicitly final; this doesn't work for anything deeper than a controller, like a service. Also, intellij doesn't quite understand them perfectly yet, so it thinks they're unused, but compilation has no errors.

PR #122, opened 2022-04-04
[PF-1448] Use Cloud kms for data encryption  

yuhuyoyo:
This PR calls GCP kms api to encrypt and decrypt the ssh key pair. A follow-on PR will handle re-encrypt the ssh key pair as kms key rotate.


service/src/main/java/bio/terra/externalcreds/dataAccess/EncryptDecryptUtils.java
```diff
@@ -0,0 +1,38 @@
...
+import com.google.protobuf.ByteString;
+import java.io.IOException;
+
+/** Utils for KMS symmetric encryption and decryption. */
+public class EncryptDecryptUtils {
+
+  private EncryptDecryptUtils() {}
+
+  /** Encrypt with KMS symmetric key. */
+  public static String encryptSymmetric(
```
dvoet: I would prefer to see this as a spring bean instead of static functions. That makes overriding and mocking much easier.

service/src/main/java/bio/terra/externalcreds/dataAccess/EncryptDecryptUtils.java
```diff
@@ -0,0 +1,38 @@
...
+      String projectId, String locationId, String keyRingId, String keyId, String plainText)
+      throws IOException {
+    try (var client = KeyManagementServiceClient.create()) {
+      var keyVersionName = CryptoKeyName.of(projectId, locationId, keyRingId, keyId);
+      EncryptResponse response = client.encrypt(keyVersionName, ByteString.copyFromUtf8(plainText));
+      client.close();
```
dvoet: isn't the close implicit in the try?

service/src/main/java/bio/terra/externalcreds/dataAccess/EncryptDecryptUtils.java
```diff
@@ -0,0 +1,38 @@
...
+
+/** Utils for KMS symmetric encryption and decryption. */
+public class EncryptDecryptUtils {
+
+  private EncryptDecryptUtils() {}
+
+  /** Encrypt with KMS symmetric key. */
+  public static String encryptSymmetric(
+      String projectId, String locationId, String keyRingId, String keyId, String plainText)
+      throws IOException {
```
dvoet: I think these functions should handle IOExceptions internally and rethrow them as ExternalCredsException. That will simplify the calling code

service/src/main/java/bio/terra/externalcreds/config/ExternalCredsConfigInterface.java
```diff
@@ -43,4 +43,17 @@
...
+    return false;
+  }
+
+  String getServiceGoogleProject();
+
+  Optional<String> getKeyRingId();
+
+  Optional<String> getKeyId();
+
+  Optional<String> getKeyRingLocation();
```
dvoet: I would have a nested config object called `KmsConfiguration` that contains non-optional fields for google project, key ring id, key id and location. Then in this object have an optional `KmsConfiguration` setting. Then you don't need the enable setting, you can be sure either all values are set or not, and the code can handle optionals more appropriately

service/src/test/java/bio/terra/externalcreds/dataAccess/SshKeyPairInternalDaoTest.java
```diff
@@ -63,6 +74,32 @@ void upsertSshKey() throws NoSuchAlgorithmException, IOException {
 
       verifySshKeyPair(sshKey, storedSshKey);
     }
+
+    @Test
+    void encryptKeyFails() {
```
dvoet: once you handle IOException internal to the encrypt and decrypt functions then the failure tests should be removed because they don't really test anything

service/src/test/java/bio/terra/externalcreds/dataAccess/SshKeyPairInternalDaoTest.java
```diff
@@ -84,6 +121,88 @@ void testGetSshKeyPair() throws NoSuchAlgorithmException, IOException {
...
+                        eq(location),
+                        eq(keyRing),
+                        eq(encryptionKey),
+                        eq(cypheredkey)))
+            .thenReturn(sshKey.getPrivateKey());
+
+        sshKeyPairDAO.upsertSshKeyPair(sshKey);
+
+        var loadedSshKeyOptional =
+            sshKeyPairDAO.getSshKeyPair(sshKey.getUserId(), sshKey.getType());
```
dvoet: I think you should verify that the database contains the cyphertext. As the test is written, you can't be certain that the encryption was actually used.

service/src/main/java/bio/terra/externalcreds/dataAccess/EncryptDecryptUtils.java
dinvlad: Yep - btw @yuhuyoyo if you look at SonarCloud code quality output, it suggests the same (along with a couple other suggestions mentioned here..)

service/src/test/java/bio/terra/externalcreds/dataAccess/SshKeyPairInternalDaoTest.java
dinvlad: Also https://sonarcloud.io/project/issues?pullRequest=122&issues=AX_12dPla-RX1g1tBSPF&open=AX_12dPla-RX1g1tBSPF&id=DataBiosphere_terra-external-credentials-manager here

service/src/test/java/bio/terra/externalcreds/dataAccess/SshKeyPairInternalDaoTest.java
```diff
@@ -84,6 +121,88 @@ void testGetSshKeyPair() throws NoSuchAlgorithmException, IOException {
...
+                        eq(externalCredsConfig.getServiceGoogleProject()),
+                        eq(location),
+                        eq(keyRing),
+                        eq(encryptionKey),
+                        eq(cypheredkey)))
+            .thenThrow(new IOException("Something went wrong, decryption fails"));
+        sshKeyPairDAO.upsertSshKeyPair(sshKey);
+        assertThrows(
+            ExternalCredsException.class,
+            () -> sshKeyPairDAO.getSshKeyPair(sshKey.getUserId(), sshKey.getType()));
```
dinvlad: Please see SonarCloud feedback about handling multiple RuntimeExceptions from this lambda and a similar one also (to be sure which function actually does throw here): https://sonarcloud.io/project/issues?resolved=false&types=CODE_SMELL&pullRequest=122&id=DataBiosphere_terra-external-credentials-manager&open=AX_12dPla-RX1g1tBSPF

dinvlad (commented):
Is re-encryption after the KMS key is rotated (by Google) in scope for this PR? As discussed earlier, we have 3 main approaches for that:

1) "Lazy" rotation upon a decryption request - detect if the KMS key used was of the "old" version, then re-encrypt with the new version and finally return the decrypted ssh key. This scales well, but might be slightly "worse" for compliance, because SSH keys that are not accessed in a while are still stored using the old versions, and we might be required to disable/destroy old versions so they would not longer be able to be decrypted then.

2) Rotate all (or a subset of) keys periodically when the old versions expire. This fully satisfies compliance I think, but has potential performance/scaling limitations. We could rotate things in stages, for example have a daily cron that re-encrypts the SSH keys that are older than 90 days on _that_ day.

3) Rotate SSH keys themselves - this also satisfies compliance (potentially even more so), but forces users to remove their old keys and upload new ones periodically from GitHub etc. every 90 days, if they wanna keep using them.

Please let me know if you have any other ideas (or if this deserves a separate PR!)

service/src/test/java/bio/terra/externalcreds/dataAccess/SshKeyPairInternalDaoTest.java
yuhuyoyo: obsolete as this test is deleted.

service/src/test/java/bio/terra/externalcreds/dataAccess/SshKeyPairInternalDaoTest.java
yuhuyoyo: i added a query to the database to check that but actually the mock should be able to check it as well as it should only return the de-cyphered text if the input matches the cyphered text.

yuhuyoyo:
> Is re-encryption after the KMS key is rotated (by Google) in scope for this PR? As discussed earlier, we have 3 main approaches for that:
> 
> 1. "Lazy" rotation upon a decryption request - detect if the KMS key used was of the "old" version, then re-encrypt with the new version and finally return the decrypted ssh key. This scales well, but might be slightly "worse" for compliance, because SSH keys that are not accessed in a while are still stored using the old versions, and we might be required to disable/destroy old versions so they would not longer be able to be decrypted then.
> 2. Rotate all (or a subset of) keys periodically when the old versions expire. This fully satisfies compliance I think, but has potential performance/scaling limitations. We could rotate things in stages, for example have a daily cron that re-encrypts the SSH keys that are older than 90 days on _that_ day.
> 3. Rotate SSH keys themselves - this also satisfies compliance (potentially even more so), but forces users to remove their old keys and upload new ones periodically from GitHub etc. every 90 days, if they wanna keep using them.
> 
> Please let me know if you have any other ideas (or if this deserves a separate PR!)

I like option 2 the best. how would we set up the cron job? I'm thinking of creating a gradle task for this and then run a Github action every evening. Any other suggestions?

Also I'd prefer to have a PR focused on one thing so i'd like to keep the key rotation implementation separate.

service/src/main/java/bio/terra/externalcreds/dataAccess/KmsEncryptDecryptHelper.java
```diff
@@ -0,0 +1,54 @@
...
+import java.io.IOException;
+import org.springframework.stereotype.Component;
+
+/** Utils for KMS symmetric encryption and decryption. */
+@Component
+public record KmsEncryptDecryptHelper(ExternalCredsConfig config) {
+
+  /** Encrypt with KMS symmetric key. */
+  public String encryptSymmetric(String plainText) {
+    try (var client = KeyManagementServiceClient.create()) {
```
dvoet: do we actually need to create a new one of these every time or can `KeyManagementServiceClient.create()` be called just once at class instantiation and that client be used for the duration?

service/src/main/java/bio/terra/externalcreds/dataAccess/SshKeyPairDAO.java
```diff
@@ -60,11 +60,15 @@ public SshKeyPairInternal upsertSshKeyPair(SshKeyPairInternal sshKeyPairInternal
             + " external_user_email = excluded.external_user_email"
             + " RETURNING id";
 
+    var sshPrivateKey = sshKeyPairInternal.getPrivateKey();
+    if (externalCredsConfig.getKmsConfiguration().isPresent()) {
+      sshPrivateKey = kmsEncryptDecryptHelper.encryptSymmetric(sshPrivateKey);
+    }
```
dvoet: 2 comments:
1) style - I would rather have this be 
```suggestion
    var sshPrivateKey;
    if (externalCredsConfig.getKmsConfiguration().isPresent()) {
      sshPrivateKey = kmsEncryptDecryptHelper.encryptSymmetric(sshKeyPairInternal.getPrivateKey());
    } else {
      sshPrivateKey = sshKeyPairInternal.getPrivateKey();
    }
```

2) This pattern means that all callers of `kmsEncryptDecryptHelper` need to do this check. Can you push it into the `kmsEncryptDecryptHelper` instead? In other words, if KMS is not configured `kmsEncryptDecryptHelper` just returns what is passed in and does not throw an exception. Maybe log a message at startup that encryption is enabled or disable.

service/src/test/resources/application-test.yaml
```diff
@@ -0,0 +1,9 @@
+spring:
+
+  datasource:
+    hikari:
+      connection-timeout: 5000
+      maximum-pool-size: 6 # cpu count * 2 https://kwahome.medium.com/database-connections-less-is-more-86c406b6fad
```
dvoet: make this a different comment that described the issue that let to this

dvoet (changes requested):

service/src/main/java/bio/terra/externalcreds/dataAccess/KmsEncryptDecryptHelper.java
yuhuyoyo: the gcp documentation suggests that `After completing all of your requests, call the "close" method on the client to
safely clean up any remaining background resources.` i imagine it's not very common that we will have a sequence of frequent encrypt and decrypt except when testing.

yuhuyoyo:
I added an expired field because when the key is previously unencrypted and we enable kms config later, we need a way to tell that the ssh key was previously un-encrypted and to be able to get the ssh key without decrypting it. This is inline with the follow-on PR where we do re-encryption of the ssh key when it expires.

dinvlad:
Makes sense to keep the rotation separate, for sure!

Re your question about cron though, Gradle job sounds like a good way to approach it. Alternatively, we could maybe add an extra endpoint to reencrypt key(s), and then call this endpoint from a Kubernetes cron job daily. Yet another alternative here is to maybe use Spring's `@Scheduled` annotation to trigger that method instead.

yuhuyoyo:
> Makes sense to keep the rotation separate, for sure!
> 
> Re your question about cron though, Gradle job sounds like a good way to approach it. Alternatively, we could maybe add an extra endpoint to reencrypt key(s), and then call this endpoint from a Kubernetes cron job daily. Yet another alternative here is to maybe use Spring's `@Scheduled` annotation to trigger that method instead.

rotation implementation is here. I will add you as a reviewer as well. https://github.com/DataBiosphere/terra-external-credentials-manager/pull/125

service/src/main/java/bio/terra/externalcreds/dataAccess/KmsEncryptDecryptHelper.java
dvoet: I looked at their examples and they are all toys. It looks to me like the client is meant for long term use. Following our use of other google client libraries, we should allocate this once.

service/src/main/resources/db/changelog/changesets/20220412_add_expires_to_ssh_key_pair_table.yaml
```diff
@@ -0,0 +1,16 @@
...
+            tableName: ssh_key_pair
+            columns:
+              - column:
+                  name: expires
+                  type: timestamp
+                  constraints:
+                    nullable: true
+                  remarks: |
+                    Timestamp when the ssh key pair encryption key is expired and the private key is
+                    due for another encryption. This column is null if the key is not encrypted.
```
dvoet: I would call this `lastEncryptedTimestamp` and don't add in the refresh duration as that is a config setting that could change. `expires` is not a good name because it sounds like the key is the thing that is expiring

dvoet (changes requested):

service/src/main/java/bio/terra/externalcreds/dataAccess/KmsEncryptDecryptHelper.java
yuhuyoyo: ok. instantiate at first method call instead of when instantiating the class as I don't want to handle api call exception in the constructor.

dvoet (approved):

PR #123, opened 2022-04-06
[PF-1384] Parse json string input to string

yuhuyoyo:
When calling this api, the request body is a json string that looks like "foo@bar.com". So we need to parse it to a string in order to remove the quotes and thus turns it into foo@bar.com.

### DIFF
--- a/service/src/main/java/bio/terra/externalcreds/controllers/SshKeyApiController.java
+++ b/service/src/main/java/bio/terra/externalcreds/controllers/SshKeyApiController.java
@@ -18,7 +18,8 @@ import org.springframework.stereotype.Controller;
 @Controller
 @Slf4j
 public record SshKeyApiController(
-    HttpServletRequest request, SamService samService, SshKeyPairService sshKeyPairService)
+    HttpServletRequest request, SamService samService, SshKeyPairService sshKeyPairService,
+    ObjectMapper objectMapper)
     implements SshKeyPairApi {
 
   @Override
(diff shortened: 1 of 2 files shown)


ddietterich (approved):
LGTM

service/src/main/java/bio/terra/externalcreds/controllers/SshKeyApiController.java
```diff
@@ -37,8 +42,15 @@ public record SshKeyApiController(
 
   @Override
   public ResponseEntity<SshKeyPair> generateSshKeyPair(SshKeyPairType type, String email) {
-    var sshKeyPair =
-        sshKeyPairService.generateSshKeyPair(getUserIdFromSam(request, samService), email, type);
-    return ResponseEntity.ok(OpenApiConverters.Output.convert(sshKeyPair));
+    try {
+      String userEmail = new ObjectMapper().readValue(email, String.class);
```
dvoet: There is a `ObjectMapper` spring bean, add it as a parameter to the `SshKeyApiController` record (see `OidcApiController`). Then use that instead of calling `new ObjectMapper()`

service/src/main/java/bio/terra/externalcreds/controllers/SshKeyApiController.java
```diff
@@ -37,8 +42,15 @@ public record SshKeyApiController(
...
+      var sshKeyPair =
+          sshKeyPairService.generateSshKeyPair(
+              getUserIdFromSam(request, samService), userEmail, type);
+      return ResponseEntity.ok(OpenApiConverters.Output.convert(sshKeyPair));
+    } catch (JsonProcessingException e) {
+      log.error("Failed to parse the email input");
```
dvoet: don't log, this is not a system error

dvoet (changes requested):

dvoet (approved):

PR #124, opened 2022-04-08
check jwt algorithm [POL-212]

Shakespeared:
Add a config to validate the JWT algorithm.

Let me know if you think there are other tests I should add, the builder complains if I set the algorithm to "none" but that should be caught by the existing test. Codecov is mad about not testing the null check but I wasn't sure how to set up a test for that since the builder complains if you try to give it improperly formed or missing headers.

### DIFF
--- a/service/src/test/java/bio/terra/externalcreds/services/PassportServiceTest.java
+++ b/service/src/test/java/bio/terra/externalcreds/services/PassportServiceTest.java
@@ -109,14 +109,12 @@ class PassportServiceTest extends BaseTest {
 
     @Test
     void testValidPassportMatchingCriteria() throws URISyntaxException {
-      when(externalCredsConfigMock.getAllowedJwtAlgorithms()).thenReturn(List.of("RS256", "ES256"));
       runValidPassportTest(new ValidPassportTestParams());
     }
 
     @Test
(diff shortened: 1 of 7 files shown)


service/src/test/java/bio/terra/externalcreds/services/PassportServiceTest.java
```diff
@@ -109,12 +109,14 @@ void testGetGA4GHPassportLinkedAccountNoPassport() {
 
     @Test
     void testValidPassportMatchingCriteria() throws URISyntaxException {
+      when(externalCredsConfigMock.getAllowedJwtAlgorithms()).thenReturn(List.of("RS256", "ES256"));
```
snwessel: Just one quick suggestion: I think that instead of mocking this in each test, you could just mock it once in the `runValidPassportTest` helper function they all call ([here](https://github.com/DataBiosphere/terra-external-credentials-manager/blob/8a4ece69617b94505e8b6c64f1e144ddd306c22b/service/src/test/java/bio/terra/externalcreds/services/PassportServiceTest.java#L207)).

dvoet (approved):

snwessel (approved):

PR #125, opened 2022-04-08
[PF-1448] set up a chronjob to check if any  SSH private keys needs re-encryption

yuhuyoyo:


yuhuyoyo:
@dvoet can I lower the test coverage bar? it's unclear to me how to add coverage for the cron application class or even if it's worth testing since the method is simply calling service.reencrypt which is fully tested in the sshkeypairservicetest.

service/src/main/java/bio/terra/externalcreds/ExternalCredsCronApplication.java
```diff
@@ -51,4 +55,11 @@ public void checkForExpiringCredentials() {
     var checkedPassportCount = providerService.validateAccessTokenVisas();
     log.info("completed validateVisas", Map.of("checked_passport_count", checkedPassportCount));
   }
+
+  @Scheduled(fixedRateString = "#{${externalcreds.background-job-interval-mins} * 60 * 1000}")
```
ddietterich, line 60: If the entire rate string could be from configuration, then you could make a test that ran the cron every second and test this dispatch code.

I am not a fan of `@Scheduled`. It is one of those Spring things is pretty, but the implementation is simple minded and not robust. By default, it runs a single thread of execution. In this code, only one of the cron tasks would run at once. Further, if the thread gets stuck or exits, no other scheduled job runs. I have debugged this case.

You can enhance Spring to do better by making a bean, but setting up the pool yourself is only a few lines of code. Debugging becomes simpler because you can easily find the thread pool.

ddietterich (approved):
Looks fine. I suggest not using `@Scheduled`, but that is probably out of scope for this PR

service/src/main/java/bio/terra/externalcreds/ExternalCredsCronApplication.java
yuhuyoyo: good to know. I added `task.scheduling.pool.size: 2` to application.yaml.

service/src/main/java/bio/terra/externalcreds/dataAccess/SshKeyPairDAO.java
```diff
@@ -91,6 +93,25 @@ public SshKeyPairInternal upsertSshKeyPair(SshKeyPairInternal sshKeyPairInternal
...
+  /** Gets expired or un-encrypted ssh key pair. */
+  public List<SshKeyPairInternal> getExpiredOrUnEncryptedSshKeyPair() {
+    var kmsConfig = externalCredsConfig.getKmsConfiguration();
+    if (kmsConfig == null) {
+      return Collections.emptyList();
+    }
+    var namedParameters =
+        new MapSqlParameterSource(
+            "refreshCutoffTimestamp",
+            Timestamp.from(Instant.now().minus(kmsConfig.getSshKeyPairRefreshDuration())));
```
dvoet: I would prefer this value to be passed in as a parameter. It is business logic to determine the cutoff. It also makes this function easier to test.

dvoet (approved):
1 little change, no re-review required

PR #126, opened 2022-04-14
PostConstruct and PreDestroy

dvoet:


service/src/main/java/bio/terra/externalcreds/dataAccess/KmsEncryptDecryptHelper.java
```diff
@@ -43,20 +44,28 @@ public String decryptSymmetric(String cypheredText) {
...
-      }
-    }
+  @PostConstruct
+  private void instantiateKeyManagementServiceClient() {
+    config
+        .getKmsConfiguration()
+        .ifPresent(
+            x -> {
+              try {
+                client = KeyManagementServiceClient.create();
```
yuhuyoyo, line 59: we would have the entire ECM service to fail to start if this fail?

PR #127, opened 2022-04-14
POL-209 preferred_username, etc.

dvoet:
a few changes:
* push requested scopes into provider config
* store redirect uri in oauth state so it does not have to be resubmitted with the oauth code
* provider config for external user id
* audit logging external user id


snwessel (approved):
Looks good to me! (I assume you see that one failing unit test though?)

PR #128, opened 2022-04-14
[POL-210] Logging transaction claims from Jwt

snwessel:
Added:
- Parsing the transaction ("txn") claim from the passport JWT
- If the transaction claim exists, then also:
  - Include it in the returned audit info
  - Include it in the audit logs when we get a new passport
- Unit tests can now specify "custom JWT claims" when generating a test passport JWT
 
Let me know if anyone can think of any other unit tests I should add.

### DIFF
--- a/service/src/main/java/bio/terra/externalcreds/services/JwtUtils.java
+++ b/service/src/main/java/bio/terra/externalcreds/services/JwtUtils.java
@@ -70,7 +70,7 @@ public record JwtUtils(ExternalCredsConfig externalCredsConfig) {
         .build();
   }
 
-  public Optional<String> getJwtTransactionId(String jwtString) {
+  public Optional<String> getJwtTransactionClaim(String jwtString) {
     var txnClaim = decodeAndValidateJwt(jwtString).getClaims().get(JWT_TRANSACTION_CLAIM);
     return Optional.ofNullable(txnClaim).map(Objects::toString);
   }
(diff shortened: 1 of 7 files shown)


service/src/main/java/bio/terra/externalcreds/services/JwtUtils.java
```diff
@@ -69,6 +70,15 @@ public PassportWithVisas decodeAndValidatePassportJwtString(String passportJwtSt
         .build();
   }
 
+  public static Optional<String> getJwtTransactionId(String jwtString) {
+    try {
+      var txnClaim = JWTParser.parse(jwtString).getJWTClaimsSet().getClaim(JWT_TRANSACTION_CLAIM);
```
dvoet: you should use `decodeAndValidateJwt`

service/src/main/java/bio/terra/externalcreds/services/PassportService.java
```diff
@@ -90,17 +91,18 @@ public ValidatePassportResultInternal validatePassport(
...
             var linkedAccount =
                 linkedAccountsByJwtId.get(passportWithVisas.getPassport().getJwtId());
+            var auditInfoMap = Map.of(
+                "passport_jti",
+                passportWithVisas.getPassport().getJwtId(),
+                "external_user_id",
+                linkedAccount.getExternalUserId(),
+                "internal_user_id",
+                linkedAccount.getUserId());
+            transactionClaim.map(t -> auditInfoMap.put("transaction_claim", t));
```
dvoet: just call it `txn`

dvoet (approved):

Shakespeared (approved):
LGTM

PR #129, opened 2022-04-15
POL-226 caching

dvoet:


service/src/main/java/bio/terra/externalcreds/ExternalCredsSpringConfig.java
```diff
@@ -19,6 +21,8 @@
 @EnableRetry
 @EnableTransactionManagement
 @EnableConfigurationProperties
+@EnableCaching
+@EnableScheduling
```
dvoet, line 25: `EnableCaching` was missing so the existing cache we had didn't even work
`EnableScheduling` moved from `ExternalCredsCronApplication` because all instances need scheduling now to handle cache resets.

snwessel (approved):

PR #132, opened 2022-05-02
[PF-1612] Move transaction to DAO and move cloud api call out of database transaction.

yuhuyoyo:


service/src/main/java/bio/terra/externalcreds/models/SshKeyPairInternal.java
```diff
@@ -1,6 +1,7 @@
 package bio.terra.externalcreds.models;
 
 import bio.terra.externalcreds.generated.model.SshKeyPairType;
+import java.sql.Timestamp;
```
ddietterich: I think using SQL timestamp is dis-recommended. It is a clumsy type. There is a simple conversion to Instant that you can make in the DAO and hide all use of the SQL type inside there . We do that uniformly in WSM, FWIW

ddietterich (approved):
LGTM - one minor suggestion about SQL timestamp type ussage

service/src/main/java/bio/terra/externalcreds/models/SshKeyPairInternal.java
yuhuyoyo: good to know. done

service/src/main/java/bio/terra/externalcreds/controllers/OpenApiConverters.java
```diff
@@ -75,11 +76,15 @@ public static LinkInfo convert(LinkedAccount linkedAccount) {
...
+    public static SshKeyPair convert(
+        SshKeyPairInternal sshKeyPairInternal, KmsEncryptDecryptHelper encryptDecryptHelper) {
       return new SshKeyPair()
           .externalUserEmail(sshKeyPairInternal.getExternalUserEmail())
           .publicKey(sshKeyPairInternal.getPublicKey())
-          .privateKey(new String(sshKeyPairInternal.getPrivateKey(), StandardCharsets.UTF_8));
+          .privateKey(
+              new String(
+                  encryptDecryptHelper.decryptSymmetric(sshKeyPairInternal.getPrivateKey()),
+                  StandardCharsets.UTF_8));
```
dvoet: all the encryption happens in the service, why is the decryption here? Or vice versa? Why not keep it all in the same place?

yuhuyoyo: yeah i tried both approaches and leave it with the minimal change to see what you would like..

if i do the decryption in the service, the sshKeyPairinternal will sometimes hold encrypted key (going in to the database) and sometimes hold decrypted key (going out).. or we can just convert to sshkeypair in the service?

service/src/main/java/bio/terra/externalcreds/dataAccess/SshKeyPairDAO.java
```diff
@@ -54,11 +64,12 @@ public boolean deleteSshKeyPairIfExists(String userId, SshKeyPairType type) {
     return jdbcTemplate.update(query, namedParameters) > 0;
   }
 
+  @WriteTransaction
   public SshKeyPairInternal upsertSshKeyPair(SshKeyPairInternal sshKeyPairInternal) {
     var query =
         "INSERT INTO ssh_key_pair (user_id, type, private_key, public_key, external_user_email, last_encrypted_timestamp)"
             + " VALUES (:userId, :type, :privateKey, :publicKey, :externalUserEmail,"
-            + (externalCredsConfig.getKmsConfiguration() != null
+            + (sshKeyPairInternal.getLastEncryptedTimestamp().isPresent()
```
dvoet: I would change this to always include `:lastEncryptedTimestamp` then set the value in `namedParameters` with
```namedParameters.addValue("lastEncryptedTimestamp", sshKeyPairInternal.getLastEncryptedTimestamp().orElse(null)```

File: service/src/main/java/bio/terra/externalcreds/services/SshKeyPairService.java [hunk 4]
```diff
@@ -32,62 +31,105 @@ public class SshKeyPairService {
...
-    return sshKeyPairDAO.upsertSshKeyPair(
-        new SshKeyPairInternal.Builder()
-            .privateKey(sshKeyPair.getPrivateKey().getBytes(StandardCharsets.UTF_8))
-            .publicKey(sshKeyPair.getPublicKey())
-            .externalUserEmail(sshKeyPair.getExternalUserEmail())
-            .userId(userId)
-            .type(type)
-            .build());
+    var privateKey = sshKeyPair.getPrivateKey().getBytes(StandardCharsets.UTF_8);
+    if (config.getKmsConfiguration() != null) {
```
dvoet: Doesn't KmsEncryptDecryptHelper do this config check? Let's try to keep this logic in one place.

service/src/main/java/bio/terra/externalcreds/services/SshKeyPairService.java
```diff
@@ -32,62 +31,105 @@ public class SshKeyPairService {
...
 
-  @WriteTransaction
+  private SshKeyPairInternal getUnencryptedSshKeyPairInternal(
+      SshKeyPairInternal sshKeyPairInternal) {
+    if (config.getKmsConfiguration() == null) {
+      return sshKeyPairInternal;
+    }
+    if (sshKeyPairInternal.getLastEncryptedTimestamp().isEmpty()) {
+      // If encryption is enabled and we are getting a previously unencrypted key, we will
+      // encrypt it as well as getting the key.
```
dvoet: I don't think it makes sense to do this. There is a background process that will take care of this. This is just added complexity for the rarest of conditions.

service/src/main/java/bio/terra/externalcreds/services/SshKeyPairService.java
```diff
@@ -97,6 +139,16 @@ public void reEncryptExpiringSshKeyPairs() {
...
+      if (sshKeyPair.getLastEncryptedTimestamp().isPresent()) {
+        encryptedKey =
+            encryptDecryptHelper.encryptSymmetric(
+                encryptDecryptHelper.decryptSymmetric(sshKeyPair.getPrivateKey()));
+      } else {
+        encryptedKey = encryptDecryptHelper.encryptSymmetric(sshKeyPair.getPrivateKey());
+      }
+      sshKeyPairDAO.upsertSshKeyPair(
+          sshKeyPair.withPrivateKey(encryptedKey).withLastEncryptedTimestamp(Instant.now()));
       sshKeyPairDAO.upsertSshKeyPair(sshKeyPair);
```
dvoet: I think you have 2 upserts here

service/src/main/java/bio/terra/externalcreds/controllers/SshKeyApiController.java
```diff
@@ -23,7 +24,8 @@ public record SshKeyApiController(
     SamService samService,
     SshKeyPairService sshKeyPairService,
     AuditLogger auditLogger,
-    ObjectMapper objectMapper)
+    ObjectMapper objectMapper,
+    KmsEncryptDecryptHelper kmsEncryptDecryptHelper)
```
dvoet: don't need this anymore

dvoet (changes requested):

dvoet (approved):

PR #133, opened 2022-05-03
POL-231 repopulate perf database before tests

dvoet:


Shakespeared (approved):

snwessel (approved):
Looks great! I'm glad to see there was a simpler way to do it than using the full bash script!

PR #135, opened 2022-06-06
POL-242 update ecm to use TCL to get sam user

dvoet:


service/src/main/java/bio/terra/externalcreds/controllers/SshKeyApiController.java
```diff
@@ -20,23 +19,24 @@
...
     AuditLogger auditLogger,
-    ObjectMapper objectMapper)
+    ObjectMapper objectMapper,
+    SamUserFactory samUserFactory)
     implements SshKeyPairApi {
 
   @Override
   public ResponseEntity<Void> deleteSshKeyPair(SshKeyPairType type) {
-    var userId = getUserIdFromSam(request, samService);
+    var samUser = samUserFactory.from(request, externalCredsConfig.getSamBasePath());
```
tlangs: I think it would be better to have a separate component that takes the `ExternalCredsConfig` in its constructor, and provides and API to just pass the request in to get the `samUser`.

tlangs (changes requested):

service/src/main/java/bio/terra/externalcreds/controllers/SshKeyApiController.java
dvoet: done

tlangs (approved):

PR #141, opened 2022-07-08
[ID-128] Change passport response to text-plain

tlangs:


dvoet (changes requested):
apis should return json format

tlangs:
This endpoint does not return JSON. It returns a base64-encoded plaintext string.

Ghost-in-a-Jar (approved):

Shakespeared (approved):

PR #142, opened 2022-07-19
ID-159 Improve ECM logging

tlangs:
K8s engines calls the status endpoint of apps several times a second, and the version endpoint once per minute. These logs are pretty useless to us, and just contribute to the amount of log messages needing to be ingested by logit and parsed through by stackdriver. It decreases our ability to diagnose problems (although we could just ignore certain messages, but we’d still be paying for them). By excluding the status and version calls from the k8s engine, we can save ~260,000 log messages a day! In each environment! For each app!!!

I also took this as an opportunity to include the stackdriver pattern layout in the logging, as well as added the ability to run with human-readable logging!


Shakespeared (approved):

service/src/main/java/bio/terra/externalcreds/logging/LoggerInterceptor.java
```diff
@@ -0,0 +1,127 @@
...
+    String responseStatus = Integer.toString(response.getStatus());
+    long requestDuration;
+    Long requestStartTime = (Long) request.getAttribute(REQUEST_START_ATTRIBUTE);
+    if (requestStartTime != null) {
+      requestDuration = System.currentTimeMillis() - requestStartTime;
+    } else {
+      requestDuration = NOT_FOUND_DURATION;
+    }
+    // skip logging the status endpoint
+    if (LOG_EXCLUDE_LIST.stream().noneMatch(url::endsWith)) {
```
andy7i: I have minor concerns about the use of `endsWith` here. If for some reason later on we wanted to add endpoints ending in `/status` or `/version` then those would also get filtered out silently.

service/src/main/java/bio/terra/externalcreds/logging/LoggerInterceptor.java
```diff
@@ -0,0 +1,127 @@
...
+        log.warn(message, stackDriverPayload);
+      } else {
+        log.info(message, stackDriverPayload);
+      }
+    } else {
+      log.debug("Received request at {}", url);
+    }
+
+    if (ex != null) {
+      log.error("An error occurred processing this request: ", ex);
```
andy7i, line 124: would it be helpful to also include `stackDriverPayload` for exceptions? Do we already get this for "free" since an exception in our code would result in a 500 error (which would be logged on L115)?

andy7i (approved):
I had a couple of questions and I'm curious what your thoughts are regarding those, but they aren't blocking. 👍

service/src/main/java/bio/terra/externalcreds/logging/LoggerInterceptor.java
tlangs: We could return 4xx responses for reasons that don't throw an exception. If a user's passport doesn't validate, we shouldn't throw an exception, but still return 400 to the user. Also, hiding exceptions in in the stackdriver payload in our logs makes them less discoverable, which is something we don't want.

andy7i: I was asking my questions to see if it would help the traceability and debug-ability of exceptions, but it sounds like it might do the opposite, in which case I'm with you

service/src/main/java/bio/terra/externalcreds/logging/LoggerInterceptor.java
tlangs: good point! changed.

andy7i: oh in that case I might suggest the same change in https://github.com/DataBiosphere/terra-drs-hub/pull/39

andy7i:
@tlangs reminder to change DR-159 to ID-159

PR #143, opened 2022-07-28
ID-159 Don't log post bodies

tlangs:
I had a realization that ECM shouldn't be logging POST bodies, as the `validatePassport` endpoint gets a passport as its body, and we shouldn't be saving that to logs! This is a follow-on to the logging improvements recently made.


Shakespeared (approved):

andy7i (approved):
maybe worth adding a comment why we aren't logging the body, but no biggie here's my 👍

tlangs:
Added a comment!

PR #144, opened 2022-08-03
ID-162 External User Passport Support

tlangs:
Adding support for the validation of passports from non-Terra users. This is a naive addition to support passports from SevenBridges, who have confirmed that the passports they send will never be more than 20 minutes old. This lets us simply check that a passport was issues less than an hour ago to validate all its visas. We don't need to reach out to RAS in this case to re-validate the visas! 

I've hardcoded the visa validity time for ease of implementation. This will probably need to be configured per-provider in the future, but since we only have RAS at this point, the hardcoded value will suffice. 

I've added tests to make sure that a "young" external passport is deemed valid, and that one over an hour old is deemed invalid.

### DIFF
--- a/service/src/main/java/bio/terra/externalcreds/services/PassportService.java
+++ b/service/src/main/java/bio/terra/externalcreds/services/PassportService.java
@@ -15,9 +15,7 @@ import bio.terra.externalcreds.models.ValidatePassportResultInternal;
 import bio.terra.externalcreds.models.VisaVerificationDetails;
 import bio.terra.externalcreds.visaComparators.VisaComparator;
 import bio.terra.externalcreds.visaComparators.VisaCriterionInternal;
-import com.nimbusds.jwt.JWTParser;
 import java.sql.Timestamp;
-import java.text.ParseException;
 import java.time.Duration;
 import java.time.Instant;
 import java.time.temporal.ChronoUnit;
@@ -144,15 +142,9 @@ public class PassportService {
   }
 
   private boolean isPassportIssueTimeValid(GA4GHPassport passport) {
-    try {
-      var issueTime = JWTParser.parse(passport.getJwt()).getJWTClaimsSet().getIssueTime().getTime();
-      var now = Instant.now().toEpochMilli();
-      return (now - issueTime < VISA_VALIDITY_TIME);
-    } catch (ParseException ex) {
-      throw new ExternalCredsException(
-          String.format("Unable to parse JWT for passport with JWT ID: %s", passport.getJwtId()),
-          ex);
-    }
+    var issueTime = jwtUtils.getJwtIssuedAt(passport.getJwt()).getTime();
+    var now = Instant.now().toEpochMilli();
+    return (now - issueTime < VISA_VALIDITY_TIME);
   }
 
   private Collection<PassportWithVisas> decodeAndValidatePassports(
--- a/service/src/main/java/bio/terra/externalcreds/services/PassportService.java
+++ b/service/src/main/java/bio/terra/externalcreds/services/PassportService.java
@@ -15,8 +15,12 @@ import bio.terra.externalcreds.models.ValidatePassportResultInternal;
 import bio.terra.externalcreds.models.VisaVerificationDetails;
 import bio.terra.externalcreds.visaComparators.VisaComparator;
 import bio.terra.externalcreds.visaComparators.VisaCriterionInternal;
+import com.nimbusds.jwt.JWTParser;
 import java.sql.Timestamp;
+import java.text.ParseException;
+import java.time.Duration;
 import java.time.Instant;
+import java.time.temporal.ChronoUnit;
 import java.util.Collection;
 import java.util.HashMap;
 import java.util.List;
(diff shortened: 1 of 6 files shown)


service/src/main/java/bio/terra/externalcreds/services/PassportService.java
```diff
@@ -37,6 +41,8 @@ public class PassportService {
   private final JwtUtils jwtUtils;
   private final Collection<VisaComparator> visaComparators;
 
+  private static final long VISA_VALIDITY_TIME = Duration.of(1, ChronoUnit.HOURS).toMillis();
```
nmalfroy: It might be easier to maintain if this is just a `Duration`

service/src/main/java/bio/terra/externalcreds/services/PassportService.java
```diff
@@ -91,34 +97,62 @@ public ValidatePassportResultInternal validatePassport(
...
             var auditInfoMap =
-                new HashMap<>(
-                    Map.of(
-                        "passport_jti",
-                        passportWithVisas.getPassport().getJwtId(),
-                        "external_user_id",
-                        linkedAccount.getExternalUserId(),
-                        "internal_user_id",
-                        linkedAccount.getUserId()));
+                new HashMap<>(Map.of("passport_jti", passportWithVisas.getPassport().getJwtId()));
```
nmalfroy, line 99: Is there a reason that you are wrapping the map in a hashmap?  Doesn't look like you ever edit it

nmalfroy: JK, just saw that you write to it.  Bummer that there's not a way to have `map.of` create a mutable map

nmalfroy (approved):
looks reasonable to me!

Shakespeared (approved):
LGTM

service/src/main/java/bio/terra/externalcreds/services/PassportService.java
tlangs: A `MutableMap.of()` feature would be lovely...

service/src/main/java/bio/terra/externalcreds/services/PassportService.java
tlangs: Then we'd be calling `toMillis()` every time we need to validate an external passport.

service/src/main/java/bio/terra/externalcreds/services/PassportService.java
```diff
@@ -91,34 +97,62 @@ public ValidatePassportResultInternal validatePassport(
...
+                                  "internal_user_id", l.getUserId()));
+                          return true;
+                        })
+                    .orElseGet(
+                        () -> {
+                          log.info(
+                              "No internal user found for passport {}. Checking issue time...",
+                              passportWithVisas.getPassport().getJwtId());
+                          return isPassportIssueTimeValid(passportWithVisas.getPassport());
+                        });
```
andy7i, line 122: this function feels like it's getting a bit unwieldy. could we pull out this logic into a helper function?

service/src/main/java/bio/terra/externalcreds/services/PassportService.java
nmalfroy: Yeah, I just feel like it's more obvious what it is if it's a duration.  At the very least, it should probably be named `VISA_VALIDITY_TIME_MILLIS` (also, I don't think the `toMillis()` call is expensive but your call

service/src/test/java/bio/terra/externalcreds/services/PassportServiceTest.java
```diff
@@ -202,7 +220,8 @@ class ValidPassportTestParams {
       String visaPhsId = "phs000123";
       String criterionPhsId = visaPhsId;
       boolean persistLinkedAccount = true;
-      Map<String, String> customJwtClaims = Collections.emptyMap();
+      boolean noLinkedAccount = false;
+      Map<String, Object> customJwtClaims = Collections.emptyMap();
```
andy7i, line 224: I haven't wrapped my head around this completely but I'm not going to have time the rest of today to continue reviewing this, so this is a bit of an incomplete thought. It feels a bit like we're testing out a test class. Should we be using an instance of whatever this class (`class ValidPassportTestParams`) is mocking out?

service/src/main/java/bio/terra/externalcreds/services/PassportService.java
tlangs: I sidestepped the issue by actually utilizing the `Duration` library to do the comparison for me!

service/src/main/java/bio/terra/externalcreds/services/PassportService.java
nmalfroy: lol, even better :)

andy7i (commented):
partial review due to meetings the rest of today

In the future, could you include the ticket link in the PR description? I think it helps me by reducing the friction and context switching cost of PR reviews!

Checking to see if I understand the requirements: it sounds like passports/visas are always considered valid within 1 hour of the issued_at time. This is true even in this extreme case where a passport is issued at 4:59pm and the user loses access a minute later at 5pm.

PR #145, opened 2022-08-23
ID-217 Make ECM scan for invalid allowlists and optionally fail-closed

tlangs:
`ncbiaccess` doesn’t fail-closed currently, and that’s a problem for us. Every time it fails, we need to go through and make sure nobody accessed files after their access was revoked. Making ECM scan for invalid allowlists and revoking access for everyone is a failsafe that will prevent unauthorized access, removing a lot of risk from the team.

We are already planning to port `ncbiaccess` functionality into to ECM. This work can act as a precursor to that porting, setting the groundwork for communication with GCS buckets and Firecloud Orchestration. The infrastructure changes needed to make this code work will also be reused for the port.


service/src/test/java/bio/terra/externalcreds/BaseTest.java
```diff
@@ -14,6 +15,12 @@
 @ActiveProfiles({"test", "human-readable-logging"})
 @Transactional
 @Rollback
+// We need this because Spring Boot doesn't re-use contexts when @MockBean is used in a test,
+// meaning a new connection pool is spun up per test class, causing us to blow through the max
+// allowable connections. Using the DirtiesContext annotation, we can tell Spring not to cache
+// the test context, meaning the connection pool will be shut down after each class, freeing up
+// connections. https://github.com/spring-projects/spring-boot/issues/7174
+@DirtiesContext(classMode = DirtiesContext.ClassMode.AFTER_CLASS)
```
tlangs, line 23: @Shakespeared This is the spring thing I was talking about that broke the test database connections!

service/src/main/resources/application.yml
```diff
@@ -15,6 +16,13 @@ externalcreds:
     key-id: ${KMS_KEY_ID:ecm-dev-kms-symmetric-key-for-ssh-key-encrypt-decrypt}
     key-ring-location: ${kMS_KEY_LOCATION:us-central1}
     ssh-key-pair-refresh-duration: ${KMS_KEY_REFRESH_DURATION:90d}
+  nih-credentials-sync-config:
+    check-frequency: ${NIH_ALLOWLIST_CHECK_CRON_STRING:-} # For once-per-hour, use @hourly
```
tlangs: `-` disables the scheduled job, so that we don't run it until the deployment shenanigans are all done.

service/analysis.gradle
```diff
@@ -24,6 +24,7 @@ spotbugs {
 spotbugsMain {
 	reports {
 		sarif { enabled = true }
+		html { enabled = true }
```
tlangs, line 27: enabling html reports because nothing can read .sarif files.

NIH_CREDS_SYNC.md
```diff
@@ -0,0 +1,22 @@
+# NIH Credentials Sync
+
+ECM is starting to port over logic from [ncbiaccess](https://github.com/broadinstitute/ncbiaccess), starting with the ability to periodically check for allowlist validity.
```
Shakespeared: should we update this description since we've changed our plans for ECM?

NIH_CREDS_SYNC.md
```diff
@@ -0,0 +1,22 @@
...
+
+## Allowlist Validity Check
+The validity check will run once-per-hour, and check that each validity list in the configured bucket is at most 12 hour old.
```
Shakespeared:
```suggestion
The validity check will run once-per-hour, and check that each validity list in the configured bucket is at most 12 hours old.
```

NIH_CREDS_SYNC.md
```diff
@@ -0,0 +1,22 @@
...
+If it detects that an allowlist is older than 12 hours, it will log errors, and if told to fail-closed, wipe out any access to the resources governed by the allowlist.
```
Shakespeared: Is the next step for this to run a manual sync if it does fail closed?

service/src/main/java/bio/terra/externalcreds/ExternalCredsCronApplication.java
```diff
@@ -71,4 +77,21 @@ public void checkForExpiringSshKeyPair() {
     }
     log.info("Completed checkForExpiringSshKeyPair");
   }
+
+  // Disabled by default in configs. Can be turned on with env var NIH_ALLOWLIST_CHECK_CRON_STRING
+  @Scheduled(cron = "${externalcreds.nih-credentials-sync-config.check-frequency}")
```
Shakespeared: I'm a little confused about the naming here, is `NIH_ALLOWLIST_CHECK_CRON_STRING` going to be the interval we want the check to run on?

service/src/main/java/bio/terra/externalcreds/services/NihCredentialsSyncService.java
```diff
@@ -0,0 +1,173 @@
...
+              } catch (Exception e) {
+                log.error(
+                    "Failed to clear access for allowlist {}. THERE IS A POTENTIAL FOR UNAUTHORIZED ACCESS",
+                    blobName,
+                    e);
+                return false;
+              }
+              return true;
+            })
+        .reduce(true, Boolean::logicalAnd);
```
Shakespeared, line 109: can you explain why it needs to `.reduce` here?

service/src/main/resources/application.yml
```diff
@@ -15,6 +16,13 @@ externalcreds:
...
+    allowlist-validity-duration: 24h
```
Shakespeared: Should we change this to slightly under 24 hours so there's a grace period?

service/src/test/java/bio/terra/externalcreds/BaseTest.java
Shakespeared: Fascinating!

Shakespeared (approved):
Left a few comments, mostly just info questions, other than that LGTM!

NIH_CREDS_SYNC.md
tlangs: Good point! Will do.

service/src/main/java/bio/terra/externalcreds/ExternalCredsCronApplication.java
tlangs: Yeah. This allows us to provide `-` to turn it off from environment variables. I agree that the env var name is a little confusing. Lemme change that.

NIH_CREDS_SYNC.md
tlangs: Good point! Will do.

service/src/main/java/bio/terra/externalcreds/services/NihCredentialsSyncService.java
tlangs: Does this clear things up? https://github.com/DataBiosphere/terra-external-credentials-manager/pull/145/files#diff-981d99066ba7ab72ba3b673eaaf350ee774d59cab8cf41599c61eb1a7d544e49R110-R113

Shakespeared: Yeah that make sense, thanks!

service/src/main/java/bio/terra/externalcreds/dataAccess/GoogleCloudStorageDAO.java
```diff
@@ -0,0 +1,122 @@
...
+    }
+  }
+
+  /**
+   * Write a {@link List<String>} to a GCS blob separated by newlines
+   *
+   * @param blobId blob id to write the lines to
+   * @param contentsToWrite contents to write to file
+   */
+  public void writeLinesToBlob(
```
gpolumbo-broad, line 61: Haven't finished the review yet, but do we need this?  I expected the logic being added here to sometimes null out the contents of a blob, but wouldn't expect it to ever write anything.

service/src/main/java/bio/terra/externalcreds/dataAccess/GoogleCloudStorageDAO.java
```diff
@@ -0,0 +1,122 @@
...
+      String googleProjectId, BlobId blobId, List<String> contentsToWrite) {
+    var newLine = "\n".getBytes(StandardCharsets.UTF_8);
+    var storage = getStorage(googleProjectId);
+    Blob blob = storage.get(blobId);
+    if (blob == null) {
+      blob = storage.create(BlobInfo.newBuilder(blobId).build());
+    }
+    try (var writer = blob.writer()) {
+      contentsToWrite.forEach(
+          s -> {
```
gpolumbo-broad, line 71: Maybe extract this function?  Yes, we can write it inline, but I personally like separating out functions, naming them appropriately, and avoiding nested try/catch blocks

service/src/main/java/bio/terra/externalcreds/dataAccess/GoogleCloudStorageDAO.java
```diff
@@ -0,0 +1,122 @@
...
+   * @param contentsToWrite contents to write to file
+   */
+  public void writeLinesToBlob(
+      String googleProjectId, BlobId blobId, List<String> contentsToWrite) {
+    var newLine = "\n".getBytes(StandardCharsets.UTF_8);
+    var storage = getStorage(googleProjectId);
+    Blob blob = storage.get(blobId);
+    if (blob == null) {
+      blob = storage.create(BlobInfo.newBuilder(blobId).build());
+    }
```
gpolumbo-broad, line 68: Nit:  I'd extract the code above this line into it's own method.  The stuff above is about getting or creating the blob.  The stuff below is about writing its contents and in addition to liking shorter methods, I also like separating logical tasks.

service/src/main/java/bio/terra/externalcreds/dataAccess/GoogleCloudStorageDAO.java
```diff
@@ -0,0 +1,122 @@
...
+              writer.write(ByteBuffer.wrap(newLine));
+            } catch (IOException e) {
+              throw new ExternalCredsException(
+                  String.format("Could not write to blob at %s. Line: %s", blobId.toGsUtilUri(), s),
+                  e);
+            }
+          });
+    } catch (IOException ex) {
+      throw new ExternalCredsException(
+          String.format("Could not write to blob at %s", blobId.toGsUtilUri()), ex);
```
gpolumbo-broad, line 83: Maybe reword this to something like, "Could not create a writer for blob %s".  I know it's a nit, but I feel like line 74 is about the actual writing and this is catching the exception for opening the writer for the blob.

service/src/main/java/bio/terra/externalcreds/dataAccess/GoogleCloudStorageDAO.java
```diff
@@ -0,0 +1,122 @@
...
+    var storage = getStorage(googleProjectId);
+    return List.of(new String(storage.readAllBytes(blobId), StandardCharsets.UTF_8).split("\n"));
+  }
+
+  // Get a Storage with application-default credentials
+  @VisibleForTesting
+  public Storage getStorage(String googleProjectId) {
+    return storageMap.computeIfAbsent(
+        googleProjectId,
+        p -> {
```
gpolumbo-broad, line 97: I am inclined to try to keep the logic inside a lambda as simple as possible.  How much is acceptable and how much is too much?  I'm not exactly sure, but this seems to fall on the "too much" side for me personally.  Multiple variable declarations, multiple steps, a try/catch, it's a lot to grok.  Not sure if this would be easier to test if it were broken or not, but I do think for maintainability, I prefer to break out logical chunks.  My take is that making something a method does not necessarily mean it needs to be reused, but that it should make the code generally more readable and maintainable as well.

service/src/main/java/bio/terra/externalcreds/services/NihCredentialsSyncService.java
```diff
@@ -0,0 +1,173 @@
...
+   * Sam permissions, which will wipe out all access for the data governed by the allowlist.
+   *
+   * @return if all allowlists are valid or not
+   */
+  public boolean allAllowlistsValid() {
+    var invalidAllowlists = getInvalidAllowlists();
+    if (invalidAllowlists.isEmpty()) {
+      log.info("No NIH allowlists are outside of their validity period");
+      return true;
+    }
```
gpolumbo-broad, line 55: More nits: Again, I would personally break this method down into multiple smaller methods.  If you would like examples, let me know.

service/src/main/java/bio/terra/externalcreds/services/NihCredentialsSyncService.java
```diff
@@ -0,0 +1,173 @@
...
+  /**
+   * Take the invalid allowlists, and update permissions so that nobody can access the data.
+   *
+   * @param invalidAllowlists allowlists that are too old to be valid anymore.
+   * @return if all the invalid allowlists could be updated and permissions revoked.
+   */
+  boolean failClosed(List<NihCredentialsBlob> invalidAllowlists) {
+    return invalidAllowlists.stream()
+        .map(
+            b -> {
```
gpolumbo-broad, line 91: Here again, I think breaking this function out would have advantages and make the lambda functionality more testable.

service/src/main/java/bio/terra/externalcreds/services/NihCredentialsSyncService.java
```diff
@@ -0,0 +1,173 @@
...
+   * @return List of blobs pointing to invalid allowlists.
+   */
+  List<NihCredentialsBlob> getInvalidAllowlists() {
+    var bucketName = externalCredsConfig.getNihCredentialsSyncConfig().getBucketName();
+    var googleProjectId = externalCredsConfig.getNihCredentialsSyncConfig().getGoogleProjectId();
+    var allowlistValidityDuration =
+        externalCredsConfig.getNihCredentialsSyncConfig().getAllowlistValidityDuration();
+    var invalidityFilter = createInvalidityFilter(Instant.now(), allowlistValidityDuration);
+
+    try (var blobsStream = googleCloudStorageDAO.streamBlobs(googleProjectId, bucketName)) {
```
gpolumbo-broad, line 129: TIL: https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html

service/src/main/java/bio/terra/externalcreds/services/NihCredentialsSyncService.java
```diff
@@ -0,0 +1,173 @@
...
+    return false;
+  }
+
+  /**
+   * Take the invalid allowlists, and update permissions so that nobody can access the data.
+   *
+   * @param invalidAllowlists allowlists that are too old to be valid anymore.
+   * @return if all the invalid allowlists could be updated and permissions revoked.
+   */
+  boolean failClosed(List<NihCredentialsBlob> invalidAllowlists) {
```
gpolumbo-broad, line 88: Not sure about current idioms in Java, but I would personally prefer to always specify the scope of a method.  I could see where an engineer changing the scope from `private` to `public` might stop and think.  But simply adding `public` might slip under the radar by accident.  Same applies for all the methods without a specified scope.

service/src/main/java/bio/terra/externalcreds/terra/FirecloudOrchestrationClient.java
```diff
@@ -0,0 +1,46 @@
...
+  private final ExternalCredsConfig externalCredsConfig;
+  private final EcmRestTemplate restTemplate;
+  private final HttpHeaders headers;
+
+  public FirecloudOrchestrationClient(
+      ExternalCredsConfig externalCredsConfig, EcmRestTemplate restTemplate) {
+    this.externalCredsConfig = externalCredsConfig;
+    this.restTemplate = restTemplate;
+    this.headers = new HttpHeaders();
+    headers.setAccept(List.of(MediaType.APPLICATION_JSON));
```
gpolumbo-broad, line 28: Can `headers` just be private static or just lazy loaded?  Looks like we aren't changing them ever.

service/src/test/java/bio/terra/externalcreds/services/NihCredentialsSyncFunctionalTest.java
```diff
@@ -0,0 +1,78 @@
...
+
+  private final Set<String> allowlistNames =
+      Set.of(
+          "test-allowlist-1.txt",
+          "test-allowlist-2.txt",
+          "test-allowlist-3.txt",
+          "phs000004.c1.allowlist.txt");
+
+  @BeforeEach
+  public void setup() {
```
gpolumbo-broad, line 46: Could we mock this out instead so that you don't have to actually read/write to google as part of the test?  You can still have a test somewhere that tests that your GoogleCloudStorageDAO can actually connect to GCP, but you don't need to test that functionality in these tests too.  

This sort of refactoring is exactly what we are going to start doing in Sam soon too, so it might be a good exercise to practice this here.

service/src/test/java/bio/terra/externalcreds/BaseTest.java
gpolumbo-broad: Could also refactor tests so that they don't actually rely on a database :)  Eventually there should be _some_ tests that actually check that we are reading, writing, and querying correctly, but we shouldn't need to do that everywhere.

gpolumbo-broad (changes requested):
In general, I am wondering whether adding all of this to ECM is the right way to go.  It was a lot of code and adding new interactions between ECM, GCS, and Orch and all the config and testing that comes with it.  Is it worth it?  How hard would this be to just tweak the existing `ncbiaccess` bash to do the same thing?  Admittedly, the logging would not be as good, but it would mean that we don't have to introduce all this stuff to ECM only to remove it when we rewrite ncbiaccess (hopefully) soon.  On the other hand, this is all already written and seems ready to go... so I don't know.

tlangs:
Closing in favor of https://github.com/broadinstitute/ncbiaccess/pull/36

PR #148, opened 2023-02-15
[PF-2478] Only run checks when PR targets default branch

dexamundsen:
Only run PR checks when the PR targets primary/default branch

currently, all PRs (irrespective of the branch they target) automatically trigger PR checks. This is unnecessary and may cause the checks to be queued up on GitHub runner.

ref -

- https://github.com/DataBiosphere/terra-workspace-manager/pull/1061
- https://broadworkbench.atlassian.net/browse/PF-2478

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -1,11 +1,13 @@
 name: Build and Test
 
 on:
+  workflow_dispatch: {}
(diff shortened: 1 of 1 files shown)


dexamundsen:
Skipping this to minimize unintended impact on repos mostly worked on by Broad team

PR #149, opened 2023-02-15
PF-2500: make private key an optional field and by default do not return it.

yuhuyoyo:
Jira: 
https://broadworkbench.atlassian.net/browse/PF-2500

What:

 Make private ssh key an optional field and only include it in the response upon request.

Why:

The UI is not using the private key field. the CLI is using the private key field. For better security, we don't include the private ssh key in the response.
 
How:

add an includPrivateKey query param with default value false.

### DIFF
--- a/common/openapi.yml
+++ b/common/openapi.yml
@@ -165,6 +165,8 @@ paths:
     parameters:
       - $ref: '#/components/parameters/sshKeyPairTypeParam'
     get:
+      parameters:
+        - $ref: '#/components/parameters/includePrivateKey'
       summary: Get the ssh key pair of the given provider.
       tags: [ ssh key pair ]
       operationId: getSshKeyPair
@@ -192,6 +194,8 @@ paths:
         '500':
           $ref: '#/components/responses/ServerError'
     post:
+      parameters:
+        - $ref: '#/components/parameters/includePrivateKey'
       summary: Generates a pair of ssh keys
       tags: [ ssh key pair ]
       operationId: generateSshKeyPair
@@ -221,7 +225,6 @@ paths:
           $ref: '#/components/responses/NotFound'
         '500':
           $ref: '#/components/responses/ServerError'
-
 components:
   parameters:
     sshKeyPairTypeParam:
@@ -230,6 +233,13 @@ components:
       schema:
         $ref: '#/components/schemas/SshKeyPairType'
       required: true
+    includePrivateKey:
+      name: includePrivateKey
+      in: query
+      required: false
+      schema:
+        type: boolean
+        default: false
     providerParam:
       name: provider
       in: path
@@ -319,7 +329,7 @@ components:
         - azure
     SshKeyPair:
       type: object
-      required: [ privateKey, publicKey, externalUserEmail ]
+      required: [ publicKey, externalUserEmail ]
       properties:
         privateKey:
           type: string
--- a/integration/src/main/java/scripts/testscripts/EncryptDecryptSshKeyPair.java
+++ b/integration/src/main/java/scripts/testscripts/EncryptDecryptSshKeyPair.java
@@ -21,16 +21,22 @@ public class EncryptDecryptSshKeyPair extends TestScript {
 
     // generate a key for each ssh key pair type
     var githubSshKeyPair =
-        sshKeyPairApi.generateSshKeyPair('"' + testUser.userEmail + '"', SshKeyPairType.GITHUB);
+        sshKeyPairApi.generateSshKeyPair(
+            '"' + testUser.userEmail + '"', SshKeyPairType.GITHUB, /*includePrivateKey=*/ true);
     var gitlabSshKeyPair =
-        sshKeyPairApi.generateSshKeyPair('"' + testUser.userEmail + '"', SshKeyPairType.GITLAB);
+        sshKeyPairApi.generateSshKeyPair(
+            '"' + testUser.userEmail + '"', SshKeyPairType.GITLAB, /*includePrivateKey=*/ false);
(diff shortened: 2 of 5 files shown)


dexamundsen (approved):

NavidZ:
What's your plan for this change rollout? If we release this change along does it not break the current integrations that do require the ssh key and still not passing the includePrivateKey parameter?

yuhuyoyo:
> What's your plan for this change rollout? If we release this change along does it not break the current integrations that do require the ssh key and still not passing the includePrivateKey parameter?

first, this is optional param, so it shouldn't break the current client.

UI doesn't use private key currently. 

Once this PR is in, I will go and update the client ECM version in CLI and UI and pass in this param. Till the ecm service with this PR rolled out to prod, we will still get private key.

dexamundsen:
> > What's your plan for this change rollout? If we release this change along does it not break the current integrations that do require the ssh key and still not passing the includePrivateKey parameter?
> 
> first, this is optional param, so it shouldn't break the current client.
> 
> UI doesn't use private key currently.
> 
> Once this PR is in, I will go and update the client ECM version in CLI and UI and pass in this param. Till the ecm service with this PR rolled out to prod, we will still get private key.

Agree
1) since clients and services are versioned, no changes are noticed in UI / CLI (others) till the client is upgraded in them. 
2) the new input param is 'optional', its default value is false. This changes the previous behavior, which is similar to the value 'true'. i.e though the API changes itself is non-breaking, the behavior is breaking. this is where the clients (cli/ui) may* break when versions are upgraded. This is the desired behavior going by the root cause of the ticket (security concerns)
3) same as (2) applies to the changes in the response object. required field is no longer required.

NavidZ:
> Once this PR is in, I will go and update the client ECM version in CLI and UI and pass in this param. Till the ecm service with this PR rolled out to prod, we will still get private key.

Axon UI only shows public key only right? So UI not asking for privatekey should be fine.
On the CLI side, the old CLI (that doesn't pass this optional parameter) reaching out to the new ECM deployment will start not getting the private key as opposed to what it used to. Is that understanding correct? Since the CLI does require it for the workspace management would that be a breaking change? 

So the question is that on the existing workspaces that have the old CLI version what happens? They will start reaching out to the new ECM but still have the old CLI installed.

yuhuyoyo:
> > Once this PR is in, I will go and update the client ECM version in CLI and UI and pass in this param. Till the ecm service with this PR rolled out to prod, we will still get private key.
> 
> Axon UI only shows public key only right? So UI not asking for privatekey should be fine. On the CLI side, the old CLI (that doesn't pass this optional parameter) reaching out to the new ECM deployment will start not getting the private key as opposed to what it used to. Is that understanding correct? Since the CLI does require it for the workspace management would that be a breaking change?
> 
> So the question is that on the existing workspaces that have the old CLI version what happens? They will start reaching out to the new ECM but still have the old CLI installed.

in old workspaces, ssh key should already be set up in existing notebook so user shouldn't need to be calling this method manually. If user created a notebook before setting up an ssh key in the UI, then unfortunately, user needs to upgrade CLI or create a new notebook.
what about I return empty string instead of null so we don't have NPE downstream?

dexamundsen:
> in old workspaces, ssh key should already be set up in existing notebook so user shouldn't need to be calling this method manually. If user created a notebook before setting up an ssh key in the UI, then unfortunately, user needs to upgrade CLI or create a new notebook. what about I return empty string instead of null so we don't have NPE downstream?

@yuhuyoyo  Can we force the CLI upgrade in notebook, say via an error message somewhere? such forced upgrades are not very uncommon. This might be a little inconvenient for some users but keeps the API clean for future use. 

on API response - sending an empty string might indicate "empty string as the key" instead of null, which indicates the intended behavior (value missing or not present)

breaking changes ought to be avoided, but would recommend cleaner APIs for easier maintenance.

yuhuyoyo:
> > in old workspaces, ssh key should already be set up in existing notebook so user shouldn't need to be calling this method manually. If user created a notebook before setting up an ssh key in the UI, then unfortunately, user needs to upgrade CLI or create a new notebook. what about I return empty string instead of null so we don't have NPE downstream?
> 
> @yuhuyoyo Can we force the CLI upgrade in notebook, say via an error message somewhere? such forced upgrades are not very uncommon. This might be a little inconvenient for some users but keeps the API clean for future use.
> 
We will have to add new code for error message or version check for the ssh command. but that requires the user to update their cLI to see :) 
i can update the CLI readme? i think the impact should be pretty small as most notebook should already have git set up and the user shouldn't need to manually add ssh key.

> on API response - sending an empty string might indicate "empty string as the key" instead of null, which indicates the intended behavior (value missing or not present)
> 
> breaking changes ought to be avoided, but would recommend cleaner APIs for easier maintenance.

i agree with you about cleaner api.

NavidZ:
The reason I brought this whole thing up was indeed to pay a special attention to backward compatibility of the APIs and the client side. If we think the risk is non-existent (at least at this point if there are no users or something) then it should be fine. But I wanna make sure we explicitly measure the risk and ensure they don't exist or have a way to remedy anything that may come up.
Also moving forward in general it is good to design APIs and client side calls in a way to allow change of behaviors in this minor behaviors. Otherwise we just have to go with introducing v2 version of the APIs (which is indeed what's used when we introduce breaking changes) and give some time to the ecosystem before deprecating v1 APIs.

Shakespeared (approved):

NavidZ (approved):

yuhuyoyo:
i checked with Willy and the Genentech folks have their ssh key set up already. so safe to make this change @NavidZ @dexamundsen

PR #156, opened 2023-09-14
[PF-2972] Update dependencies

zloery:
Jira: [PF-2972](https://broadworkbench.atlassian.net/browse/PF-2972)

What:
Update a number of dependencies with known vulnerabilities.

Why:
Verily security has started using a different dependency analysis tool which flagged a handful of vulnerabilities. I'd like to update dependencies to fix issues raised by both Verily's tool and Sourceclear.

How:
I verified the dependency versions are changed by generating lockfiles locally, though it looks like this service generally isn't using them anymore so I didn't include them in this PR.
  

[PF-2972]: https://broadworkbench.atlassian.net/browse/PF-2972?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/client-resttemplate/swagger.gradle
+++ b/client-resttemplate/swagger.gradle
@@ -4,6 +4,13 @@ dependencies {
 
 	implementation 'io.swagger.core.v3:swagger-annotations:2.1.13'
 	swaggerCodegen 'io.swagger.codegen.v3:swagger-codegen-cli:3.0.33'
+
+	// Transitive dependency constraints due to security vulnerabilities in prior versions.
+	// These are not directly included, they are just constrained if they are pulled in as
+	// transitive dependencies.
+	constraints {
+		implementation('org.yaml:snakeyaml:1.31')
+	}
 }
 
 
--- a/build.gradle
+++ b/build.gradle
@@ -14,6 +14,11 @@ plugins {
 
 repositories { mavenCentral() }
 
+// Dependency versions for libraries managed by Spring dependency management, affecting all projects
+// This can be removed once ECM has migrated to Spring Boot 3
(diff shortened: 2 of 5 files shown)


client-resttemplate/swagger.gradle
```diff
@@ -4,6 +4,13 @@ dependencies {
 
 	implementation 'io.swagger.core.v3:swagger-annotations:2.1.13'
 	swaggerCodegen 'io.swagger.codegen.v3:swagger-codegen-cli:3.0.33'
+
+	// Transitive dependency constraints due to security vulnerabilities in prior versions.
+	// These are not directly included, they are just constrained if they are pulled in as
+	// transitive dependencies.
+	constraints {
+		implementation('org.yaml:snakeyaml:1.31')
```
Shakespeared, line 12: Last time I did deps updates sourceclear flagged snakeYAML versions under 2.0 as unsafe and there was a whole rigamarole with snakeYAML 2.0 requiring spring boot 3 which broke a bunch of other things. Does 1.31 fix the vulnerability without requiring a spring boot upgrade?

zloery: Unfortunately there are still issues with the latest 1.X version that are only fixed in 2.0 and beyond, and they still need the spring boot 3 upgrade. 
However, there are other issues with ECM's current version (1.30) that are fixed in 1.31 (and given that 1.33 is also compatible, I've upgraded a bit farther)

Shakespeared (approved):

PR #160, opened 2023-12-13
[DR-3397] ECM publishes Jakarta-based client

okotsopoulos:
**Jira:** https://broadworkbench.atlassian.net/browse/DR-3397

**What:**

ECM publishes a Jakarta-based client library at `bio.terra:externalcreds-client-resttemplate-jakarta` while still maintaining its existing Javax-based client library at `bio.terra:externalcreds-client-resttemplate`.

**Why:**

This gives clients running Spring 3 (imminently, DrsHub) a compatible Client Library.  More info: https://github.com/DataBiosphere/terra-drs-hub/pull/97/files#r1424657737

**How:**

Following existing precedent set by @dvoet in https://github.com/DataBiosphere/jade-data-repo/pull/1551 for Datarepo -- thanks, Doug!

**Verification**

I published this Jakarta-based client locally, pulled it into my local DrsHub, and observed that DRS URI resolution which previously failed when communicating with ECM due to client incompatibilities now succeeds.


dvoet (approved):
the changes look good but I don't know if we need to keep the javax version. Are there clients besides TDR? The API has not been changing much so clients doin't need to update anyway.

okotsopoulos:
@dvoet , good point -- I know that both TDR and DrsHub use the ECM client.  DrsHub is getting upgraded to Spring 3, but TDR is still on Spring 2 for the time being and likely won't get upgraded until the new year (it's a more extensive nest of upgrades compared to DrsHub).

Ghost-in-a-Jar (approved):

Shakespeared (approved):

PR #161, opened 2023-12-14
[DR-3398] [ID-948] Upgrade to Spring Boot 3.1.6, TCL 0.1.9

okotsopoulos:
**Jira:**
https://broadworkbench.atlassian.net/browse/DR-3398
https://broadworkbench.atlassian.net/browse/ID-948

**What:**

In a major version release:
- Upgrade ECM to Spring Boot 3.1.6, Terra Common Lib 0.1.9
- Move back to only publishing a single client library, now Jakarta-based: `bio.terra:externalcreds-client-resttemplate`
  - This means that any service who wants the latest ECM client library will also need to move to Spring Boot 3. From what I can see, that's only DrsHub (migration in progress), TDR (migration coming soon), and Terra CLI (no longer under our ownership) so that seems acceptable.

**Why:**

Yesterday I merged a PR to make ECM publish a Jakarta-based client library in addition to the existing Javax-based client library.  At that point I didn't pursue a full upgrade to Spring Boot 3: https://github.com/DataBiosphere/terra-external-credentials-manager/pull/160

Initially, testing the new client library with DrsHub on Spring Boot 3 seemed to work, but I was only able to test DRS URI resolution for a passportable snapshot when the caller held no passport.

After merging, I was able to [re-link my RAS account in Terra Dev](https://broadinstitute.slack.com/archives/C03GMG4DUSE/p1702493607366309?thread_ts=1702487011.872899&cid=C03GMG4DUSE) and verified that I now held a passport.  When retesting DRS URI resolution for a passportable snapshot, I was running into client incompatibility errors similar to the ones I saw before.

Instead of figuring out how to resolve issues specific to the second client library, I instead decided that it would be more worthwhile to pursue the upgrade and move back to a single client library.  Publishing the new library locally and pulling it into my local DrsHub now allows DRS URI resolution for a passportable snapshot to proceed when the caller holds a passport.

Side benefits: these upgrades allowed me to clean up package pinning related to past security vulnerabilities, and more completely leverage the dependency management plug-in.  Identiteam may like to cross-check any open security vulnerability tickets on their board for ECM to see if this PR resolves them.
- ETA: this PR addresses https://nvd.nist.gov/vuln/detail/cve-2023-6378 by pinning `logback-classic`, `logback-core` to 1.4.14

**How:**

Following precedence set by prior Spring 3 upgrades and https://broadworkbench.atlassian.net/wiki/spaces/DSPBLOG/blog/2023/11/16/2906521657/The+Great+Spring+Boot+3+Yak+Shaving+Expedition (thanks, @dvoet!)

To review, you may prefer to go commit-by-commit: I endeavored to split them up logically (e.g. the "javax -> jakarta" commit touches many files but they're all package name changes).


service/build.gradle
```diff
@@ -53,6 +53,18 @@ dependencies {
 	}
 	// https://mvnrepository.com/artifact/org.mockito/mockito-core
 	testImplementation group: 'org.mockito', name: 'mockito-core', version: '4.4.0'
+
+	// OpenTelemetry
+	var openTelemetryVersion = '1.31.0'
```
dvoet:
```suggestion
	var openTelemetryVersion = '1.32.0'
```

service/src/main/resources/application.yml
```diff
@@ -20,6 +20,19 @@ logging:
   level.bio.terra.externalcreds: ${EXTERNALCREDS_LOG_LEVEL:debug}
   pattern.level: '%X{requestId} %5p'
 
+otel:
+  sdk:
+    disabled: false # set to true to disable all open telemetry features
+  traces:
+    sampler:
+      probability: ${SAMPLING_PROBABILITY:0}
```
dvoet: this was incorrect in earlier iterations, sampling is set [like this](https://github.com/DataBiosphere/terra-workspace-manager/blob/3958b8cb975444950305a74a126d59fcc9974136/service/src/main/resources/application.yml#L182)

client-resttemplate/swagger.gradle
```diff
@@ -21,7 +24,8 @@ generateSwaggerCode {
 			modelPackage  : "${artifactGroup}.model",
 			apiPackage    : "${artifactGroup}.api",
 			invokerPackage: "${artifactGroup}.client",
-			dateLibrary   : 'java11'
+			dateLibrary   : 'java11',
+			jakarta       : true
```
dvoet, line 28: I didn't say this in the last review but it has been nagging me. It might be worth switching from `library = 'resttemplate'` above to `library = 'jersey2'` to be like other service. Then cross service tracing can be [setup like this](https://github.com/DataBiosphere/terra-workspace-manager/blob/3958b8cb975444950305a74a126d59fcc9974136/service/src/main/java/bio/terra/workspace/service/policy/TpsApiDispatch.java#L65). Alternatively we could implement a `ClientHttpRequestInterceptor` that does tracing for this kind of client lib.

dvoet (approved):
A couple minor things to address and one question to consider. Thanks for taking this work forward.

Ghost-in-a-Jar (approved):

service/build.gradle
```diff
@@ -28,14 +28,14 @@ apply from: 'analysis.gradle'
...
-	implementation 'org.bouncycastle:bcprov-jdk15on:1.70'
-	implementation 'org.codehaus.janino:janino:3.1.7' // Provides if/else xml parsing for logback config
+	annotationProcessor 'org.immutables:value:2.9.2'
+	implementation 'bio.terra:terra-common-lib:0.1.9-SNAPSHOT'
+	// https://mvnrepository.com/artifact/org.bouncycastle/bcprov-jdk18on
+	implementation 'org.bouncycastle:bcprov-jdk18on:1.76'
+	implementation 'org.codehaus.janino:janino:3.1.9' // Provides if/else xml parsing for logback config
 	implementation('org.springframework.cloud:spring-cloud-gcp-starter-logging:1.2.8.RELEASE') {
 		exclude group: 'org.springframework', module: 'spring-jcl'
 	}
```
pshapiro4broad: With the latest dependency plugin, is this pinned version and exclude still required?

service/build.gradle
pshapiro4broad: After seeing this in different PRs I'm wondering if we should expose this dependency from tcl, with the idea that all terra services should be using tracing and opentelemetry.

okotsopoulos: I'm glad you brought this up, I was wondering about this too.  @dvoet , is there a reason why this block isn't handled by default in TCL?

client-resttemplate/swagger.gradle
okotsopoulos: Good shout, I can play with this.

Do you have a sense of what changes would be required to the services interacting with the client library?

service/build.gradle
okotsopoulos: It is not!  I also changed the parent package following this morning's discussion: `org.springframework.cloud:spring-cloud-gcp-starter-logging` is no longer maintained, we get this from `com.google.cloud` now.

Shakespeared (approved):
ltgm

client-resttemplate/swagger.gradle
okotsopoulos: I think I'd like to work on this separately in its own ticket, so will merge this PR in without it… I couldn't get it to work out of the box but I may be missing something simple.

PR #164, opened 2024-02-07
Separate passport-provider specific code

samanehsan:
**Ticekt:**
This will make it easier to implement https://broadworkbench.atlassian.net/browse/ID-1051 (I can create a separate ticket for this refactor if that makes code review easier).

**What:**
Refactor the OidcApiController `createLink` method and `ProviderService` to separate out the passport provider-specific logic.

**Why:**
We want to add support for linking a GitHub account via ECM, and the functionality to exchange an oauth code for a refresh token has already been implemented. However, it is specific to linking a Ras provider so it is not currently reusable.

**How:**
- Move passport-related methods from `ProviderService` to the new `PassportProviderService`.
- Generalize `createLinkInternal` so it can be re-used for other provider types
- Move `createLink` into PassportProviderService (token and fence account services will have different implementations)
- Move passport-specific audit logging out of OidcApiController `createLink`
- Use an enum for the provider parameter instead of an arbitrary string

### DIFF
--- a/service/src/main/java/bio/terra/externalcreds/services/ProviderService.java
+++ b/service/src/main/java/bio/terra/externalcreds/services/ProviderService.java
@@ -129,7 +129,7 @@ public class ProviderService {
     }
   }
 
-  public ImmutablePair<LinkedAccount, OAuth2User> createLinkInternal(
+  protected ImmutablePair<LinkedAccount, OAuth2User> createLinkedAccount(
       String providerName,
       String userId,
       String authorizationCode,
@@ -183,7 +183,7 @@ public class ProviderService {
    * OAuth2AuthorizationExceptions with other OAuth2AuthorizationExceptions that have non-standard
    * error codes. We just want to handle standard error codes which should be in the root cause.
    */
-  public String getRootOAuth2ErrorCode(OAuth2AuthorizationException oauthEx) {
+  protected String getRootOAuth2ErrorCode(OAuth2AuthorizationException oauthEx) {
     var errorCode = oauthEx.getError().getErrorCode();
     Throwable currentThrowable = oauthEx.getCause();
     var visitedThrowables = new ArrayList<Throwable>();
@@ -219,7 +219,7 @@ public class ProviderService {
     return linkedAccount;
   }
 
-  public void invalidateLinkedAccount(LinkedAccount linkedAccount) {
+  protected void invalidateLinkedAccount(LinkedAccount linkedAccount) {
     auditLogger.logEvent(
         new AuditLogEvent.Builder()
             .auditLogEventType(AuditLogEventType.LinkExpired)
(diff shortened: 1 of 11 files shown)


service/src/main/java/bio/terra/externalcreds/services/ProviderService.java
```diff
@@ -186,7 +129,7 @@ private OAuth2State validateOAuth2State(String providerName, String userId, Stri
     }
   }
 
-  private LinkedAccountWithPassportAndVisas createLinkInternal(
+  public ImmutablePair<LinkedAccount, OAuth2User> createLinkInternal(
```
tlangs: can we rename this? Since its a public method, its not really internal anymore.

tlangs: Alternatively, we can make it `protected`, since its only used by the subclass(es)

service/src/main/java/bio/terra/externalcreds/services/ProviderService.java
```diff
@@ -335,7 +182,7 @@ void authAndRefreshPassport(LinkedAccount linkedAccount) {
    * OAuth2AuthorizationExceptions with other OAuth2AuthorizationExceptions that have non-standard
    * error codes. We just want to handle standard error codes which should be in the root cause.
    */
-  private String getRootOAuth2ErrorCode(OAuth2AuthorizationException oauthEx) {
+  public String getRootOAuth2ErrorCode(OAuth2AuthorizationException oauthEx) {
```
tlangs: can be `protected`

service/src/main/java/bio/terra/externalcreds/services/ProviderService.java
```diff
@@ -352,81 +199,28 @@ private String getRootOAuth2ErrorCode(OAuth2AuthorizationException oauthEx) {
...
-    if (visaValid) {
-      passportService.updateVisaLastValidated(visaDetails.getVisaId());
-    }
-    return visaValid;
+    return linkedAccount;
   }
 
-  private void invalidateLinkedAccount(LinkedAccount linkedAccount) {
+
+  public void invalidateLinkedAccount(LinkedAccount linkedAccount) {
```
tlangs: can also be `protected`

tlangs (changes requested):

service/src/test/java/bio/terra/externalcreds/TestUtils.java
```diff
@@ -20,9 +20,13 @@ public static Timestamp getRandomTimestamp() {
   }
 
   public static LinkedAccount createRandomLinkedAccount() {
+    return createRandomLinkedAccount("ras");
```
samanehsan: I want to change the provider param from a string type to an enum and replace this hard-coded value: https://github.com/DataBiosphere/terra-external-credentials-manager/blob/dev/common/openapi.yml#L247

tlangs: That sounds like an excellent idea!

service/src/main/java/bio/terra/externalcreds/services/ProviderService.java
samanehsan: I can do both! What do you think about renaming this to `createLinkedAccount`? 🤔

tlangs: Sounds great!

tlangs (approved):

Ghost-in-a-Jar (approved):

PR #165, opened 2024-02-12
[ID-1051] Add GitHub account linking

samanehsan:
**Jira:** 
- https://broadworkbench.atlassian.net/browse/ID-1054
- https://broadworkbench.atlassian.net/browse/ID-1050
- https://broadworkbench.atlassian.net/browse/ID-1051

**What:**
Add `github` as a provider option for the following endpoints:
- GET `/api/oidc/v1/{provider}`
- DELETE `/api/oidc/v1/{provider}`
- GET `/api/oidc/v1/{provider}/authorization-url`
- POST `/api/oidc/v1/{provider}/oauthcode`

The only endpoint changed in this PR is the POST endpoint. If the provider is `github`, the createLink method for token providers will be called (not passport providers).

**Why:**
This PR enables a user to link their github account, which will allow them to import private workflows into their Terra workspace.

**How:**
- Add a new `TokenProviderService` with a `createLink` method that handles account linking for any provider that uses access tokens. 

Note: This is blocked by https://broadworkbench.atlassian.net/browse/ID-1044 (adding github to the ECM provider configuration)

### DIFF
--- a/service/src/main/java/bio/terra/externalcreds/controllers/OidcApiController.java
+++ b/service/src/main/java/bio/terra/externalcreds/controllers/OidcApiController.java
@@ -7,6 +7,7 @@ import bio.terra.externalcreds.auditLogging.AuditLogger;
 import bio.terra.externalcreds.controllers.OpenApiConverters.Output;
 import bio.terra.externalcreds.generated.api.OidcApi;
 import bio.terra.externalcreds.generated.model.LinkInfo;
+import bio.terra.externalcreds.generated.model.PassportProvider;
 import bio.terra.externalcreds.generated.model.Provider;
 import bio.terra.externalcreds.models.LinkedAccount;
 import bio.terra.externalcreds.services.JwtUtils;
@@ -130,7 +131,7 @@ public record OidcApiController(
   }
 
   @Override
-  public ResponseEntity<String> getProviderPassport(Provider providerName) {
+  public ResponseEntity<String> getProviderPassport(PassportProvider providerName) {
     var samUser = samUserFactory.from(request);
     var maybeLinkedAccount =
         linkedAccountService.getLinkedAccount(samUser.getSubjectId(), providerName.name());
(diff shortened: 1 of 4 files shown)


service/src/main/java/bio/terra/externalcreds/services/TokenProviderService.java
```diff
@@ -0,0 +1,82 @@
...
+        externalCredsConfig,
+        providerClientCache,
+        oAuth2Service,
+        linkedAccountService,
+        auditLogger,
+        objectMapper);
+  }
+
+  public Optional<LinkedAccount> createLink(
+      String providerName,
```
tlangs, line 37: This should be a `Provider`, not a `String`. That way, we can do validation on it.

service/src/main/java/bio/terra/externalcreds/controllers/OidcApiController.java
```diff
@@ -76,6 +80,7 @@ public ResponseEntity<LinkInfo> createLink(
             .userId(samUser.getSubjectId())
             .clientIP(request.getRemoteAddr());
 
+    Optional<LinkInfo> linkInfo;
     try {
       if (providerName.equals(Provider.RAS)) {
```
tlangs, line 85: Since `providerName` is an enum value, we can replace this `if/else` with a `switch` statement

tlangs (changes requested):
We should try and use the `Provider` enum as much as we can in the codebase. Passing magic strings around is bound to lead to confusion if we keep adding providers.

service/src/main/java/bio/terra/externalcreds/services/TokenProviderService.java
samanehsan: I'll change it to be consistent with https://github.com/DataBiosphere/terra-external-credentials-manager/pull/168. I'm not sure that we need to check if this is a valid provider here since that already happens upstream. Unless you mean validating that it is a token-based provider?

PR #168, opened 2024-02-13
[ID-1085] Fix provider lookup

samanehsan:
**Jira:** https://broadworkbench.atlassian.net/browse/ID-1085

**What/Why/How:**
Fixes an issue where the OIDC endpoints throw a 404 error because they are using the provider `name` which is in all-caps, but the name is lowercase in the provider configuration. This PR changes this to use the string value of the enum instead.

### DIFF
--- a/service/src/main/java/bio/terra/externalcreds/controllers/OidcApiController.java
+++ b/service/src/main/java/bio/terra/externalcreds/controllers/OidcApiController.java
@@ -49,7 +49,8 @@ public record OidcApiController(
   @Override
   public ResponseEntity<LinkInfo> getLink(Provider providerName) {
     var samUser = samUserFactory.from(request);
-    var linkedAccount = linkedAccountService.getLinkedAccount(samUser.getSubjectId(), providerName);
+    var linkedAccount =
+        linkedAccountService.getLinkedAccount(samUser.getSubjectId(), providerName.toString());
     return ResponseEntity.of(linkedAccount.map(OpenApiConverters.Output::convert));
   }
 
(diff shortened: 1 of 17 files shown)


service/src/main/java/bio/terra/externalcreds/controllers/OidcApiController.java
```diff
@@ -50,7 +50,7 @@ public ResponseEntity<List<String>> listProviders() {
   public ResponseEntity<LinkInfo> getLink(Provider providerName) {
     var samUser = samUserFactory.from(request);
     var linkedAccount =
-        linkedAccountService.getLinkedAccount(samUser.getSubjectId(), providerName.name());
+        linkedAccountService.getLinkedAccount(samUser.getSubjectId(), providerName.toString());
```
tlangs, line 53: we should probably just pass the entire `Provider` here, and let whatever needs the string value do what it needs to do.

tlangs (changes requested):

service/src/main/java/bio/terra/externalcreds/controllers/OidcApiController.java
samanehsan: That's a good point! I initially did it this way because it's used as a string in a lot of places 😕

samanehsan: So I started making this change and it impacts a ton of places in the code. And in some tests we would need to convert the provider name string (stored on linked account objects) to an enum so...I'm not convinced it's worth making this change 🙈. What do you think about leaving it as-is?

tlangs: Sure, lets just leave it for now, but can we make a ticket to migrate to using the enum?

samanehsan: Oh I just saw your comment! I'm nearly done with the change but I'm running into an issue with the tests so I'll put it in a separate PR so this isn't blocked.

Ghost-in-a-Jar (approved):

samanehsan:
Closing in favor of https://github.com/DataBiosphere/terra-external-credentials-manager/pull/169 which uses the correct ticket number in the branch name

PR #169, opened 2024-02-13
[ID-1085] Fix provider lookup

samanehsan:
**Jira:** https://broadworkbench.atlassian.net/browse/ID-1085

**What/Why/How:**
Fixes an issue where the OIDC endpoints throw a 404 error because they are using the provider name which is in all-caps, but the name is lowercase in the provider configuration. This PR changes this to use the string value of the enum instead.
  
(This PR has the correct ticket number in the branch name)


samanehsan:
This is the same PR as https://github.com/DataBiosphere/terra-external-credentials-manager/pull/168 (which has two approvals) with the correct branch name

PR #170, opened 2024-02-15
[ID-1054] Return GitHub auth url

samanehsan:
**Jira:** https://broadworkbench.atlassian.net/browse/ID-1054

**What:**
When a user calls `/api/v1/oidc/github/authorization-url` with an allowed `redirectUri`, they should get an authorization url in the format https://github.com/login/oauth/authorize?response_type=code&client_id={id}&state={state}&redirect_uri={redirect_uri}. When you go to that authorization url, it will take you to the github authorization page. When you agree to link your account it will redirect you to the `redirectUri` (terra dev) and add a `code` query parameter to the url.

This PR also deprecates `/api/oidc/v1/github/authorization-url` in favor of `/api/oauth/v1/github/authorization-url` (the only difference is the route name change from oidc -> oauth).
Screenshot 2024-02-16 at 11 55 30 AM

**Why:**
This is the first step in initiating the authorization flow with github. The Terra UI will: 
1. Call GET `/api/oauth/v1/github/authorization-url`
2. Call the github authorization url, which will prompt the user to authorize their account in github and then redirect to the Terra UI
3. Call the POST `/api/oauth/v1/github/oauthcode` endpoint with the state and code (this will exchange the oauth code for a refresh token and store a record of the user's linked account in ECM)

**How:**
- Add `github` as another provider option to all OIDC API endpoints (except for GET /passports)
- Update how the `ClientRegistration` is created to work for Github. The existing implementation was specific to OIDC providers. 

**Testing:**
1. Run ECM locally and authenticate as a Terra dev user
2. Call `/api/oauth/v1/github/authorization-url` with the callback url "https://bvdp-saturn-dev.appspot.com/oauth_callback"
3. Navigate to the url returned by that endpoint and authorize the Terra github app
4. When you get redirected to Terra, confirm that there is now a "code" query parameter in the url

### DIFF
--- a/common/openapi.yml
+++ b/common/openapi.yml
@@ -93,7 +93,7 @@ paths:
 
   /api/oidc/v1/{provider}/passport:
     parameters:
-      - $ref: '#/components/parameters/providerParam'
+      - $ref: '#/components/parameters/passportProviderParam'
     get:
       summary: Get the original passport from this provider
       tags: [ oidc ]
@@ -240,6 +240,12 @@ components:
       schema:
         type: boolean
         default: false
+    passportProviderParam:
+      name: provider
+      in: path
+      required: true
+      schema:
+        $ref: '#/components/schemas/PassportProvider'
     providerParam:
       name: provider
       in: path
@@ -319,11 +325,17 @@ components:
             $ref: '#/components/schemas/ErrorReport'
 
   schemas:
+    PassportProvider:
+      description: Enum containing valid passport providers.
+      type: string
+      enum:
+        - ras
     Provider:
       description: Enum containing valid identity providers.
       type: string
       enum:
         - ras
+        - github
     SshKeyPairType:
       description: Enum containing valid ssh key types.
       type: string
--- a/common/openapi.yml
+++ b/common/openapi.yml
@@ -4,6 +4,24 @@ info:
   description: A manager for external credentials
   version: 0.0.1
 paths:
+  /api/oauth/v1/{provider}/authorization-url:
+    parameters:
+      - $ref: '#/components/parameters/providerParam'
+      - $ref: '#/components/parameters/redirectUriParam'
+    get:
+      summary: Builds an OAuth authorization URL that a user must use to initiate the OAuth dance.
+      tags: [ oauth ]
+      operationId: getAuthUrl
+      description: First step to link a new provider. Creates a 1 time use link. There can only be 1 active link per user per provider. The last link created will invalidate any prior links.
+      responses:
+        '200':
+          description: The authorization URL
+          content:
+            application/json:
+              schema:
+                type: string
+        '500':
+          $ref: '#/components/responses/ServerError'
   /api/oidc/v1/providers:
     get:
       summary: Lists the available OIDC providers.
@@ -55,6 +73,7 @@ paths:
     get:
       summary: Builds an OAuth authorization URL that a user must use to initiate the OAuth dance.
       tags: [ oidc ]
+      deprecated: true
       operationId: getAuthUrl
       description: First step to link a new provider. Creates a 1 time use link. There can only be 1 active link per user per provider. The last link created will invalidate any prior links.
       responses:
--- a/common/openapi.yml
+++ b/common/openapi.yml
@@ -11,13 +11,13 @@ paths:
     get:
       summary: Builds an OAuth authorization URL that a user must use to initiate the OAuth dance.
       tags: [ oauth ]
-      operationId: getAuthUrl
+      operationId: getAuthorizationUrl
       description: First step to link a new provider. Creates a 1 time use link. There can only be 1 active link per user per provider. The last link created will invalidate any prior links.
       responses:
         '200':
           description: The authorization URL
           content:
-            application/json:
+            text/plain:
               schema:
                 type: string
         '500':
--- a/service/src/main/java/bio/terra/externalcreds/config/ProviderPropertiesInterface.java
+++ b/service/src/main/java/bio/terra/externalcreds/config/ProviderPropertiesInterface.java
@@ -28,6 +28,8 @@ public interface ProviderPropertiesInterface {
 
   String getExternalIdClaim();
 
+  String getUserNameAttributeName();
+
   // optional overrides for values in provider's /.well-known/openid-configuration
   Optional<String> getUserInfoEndpoint();
 
--- a/service/src/main/java/bio/terra/externalcreds/controllers/OauthApiController.java
+++ b/service/src/main/java/bio/terra/externalcreds/controllers/OauthApiController.java
@@ -1,0 +1,37 @@
+package bio.terra.externalcreds.controllers;
+
+import bio.terra.externalcreds.generated.api.OauthApi;
+import bio.terra.externalcreds.generated.model.Provider;
+import bio.terra.externalcreds.services.*;
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import jakarta.servlet.http.HttpServletRequest;
+import lombok.SneakyThrows;
+import org.springframework.http.ResponseEntity;
+import org.springframework.stereotype.Controller;
+
+@Controller
+public record OauthApiController(
+    HttpServletRequest request,
+    ObjectMapper mapper,
+    ProviderService providerService,
+    ExternalCredsSamUserFactory samUserFactory)
+    implements OauthApi {
+
+  @Override
+  public ResponseEntity<String> getAuthUrl(Provider providerName, String redirectUri) {
+    var samUser = samUserFactory.from(request);
+
+    var authorizationUrl =
+        providerService.getProviderAuthorizationUrl(
+            samUser.getSubjectId(), providerName.toString(), redirectUri);
+
+    return ResponseEntity.of(authorizationUrl.map(this::jsonString));
+  }
+
+  /** Helper method to format a string as json. Otherwise it isn't quoted or escaped correctly. */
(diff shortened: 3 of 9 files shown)


service/src/main/java/bio/terra/externalcreds/services/ProviderClientCache.java
```diff
@@ -41,12 +45,32 @@ public void resetCache() {
...
+      String providerName, ProviderProperties providerInfo) {
+    ClientRegistration.Builder builder;
+    if (providerName.equals(Provider.RAS.toString())) {
+      builder =
+          ClientRegistrations.fromOidcIssuerLocation(providerInfo.getIssuer())
+              .clientId(providerInfo.getClientId())
+              .clientSecret(providerInfo.getClientSecret())
+              .issuerUri(providerInfo.getIssuer());
+    } else if (providerName.equals(Provider.GITHUB.toString())) {
+      var redirectUri =
```
samanehsan: Can a provider have multiple valid redirect uris? (Right now RAS and Github each have one valid redirect uri). I'm wondering if I should pass in the redirect uri from the API request or if it is okay to get it from the config.

Ghost-in-a-Jar: You can configure multiple "Callback URL"'s in the github app but not sure why we would need more than one, I would think getting it from either source would be fine. Maybe getting it from the request makes sense just so if we change it we have to update it in less places? I dont think itll make much of a difference either way though.

service/src/main/java/bio/terra/externalcreds/services/ProviderClientCache.java
```diff
@@ -41,12 +45,32 @@ public void resetCache() {
...
-  private ClientRegistration buildClientRegistration(ProviderProperties providerInfo) {
-    var builder =
-        ClientRegistrations.fromOidcIssuerLocation(providerInfo.getIssuer())
-            .clientId(providerInfo.getClientId())
-            .clientSecret(providerInfo.getClientSecret())
-            .issuerUri(providerInfo.getIssuer());
+  private ClientRegistration buildClientRegistration(
+      String providerName, ProviderProperties providerInfo) {
+    ClientRegistration.Builder builder;
+    if (providerName.equals(Provider.RAS.toString())) {
```
Ghost-in-a-Jar: This is just a suggestion, feel free to ignore, but would this if/else if/else block be more succinct as a pattern matching switch statement (or just a regular switch statment)? Not sure which version of java ECM uses https://docs.oracle.com/en/java/javase/17/language/pattern-matching-switch-expressions-and-statements.html#GUID-E69EEA63-E204-41B4-AA7F-D58B26A3B232
My rationale is because we will probably add more providers in the future.

service/src/main/java/bio/terra/externalcreds/services/ProviderClientCache.java
```diff
@@ -41,12 +45,32 @@ public void resetCache() {
     log.info("ProviderClientCache reset");
   }
 
-  private ClientRegistration buildClientRegistration(ProviderProperties providerInfo) {
-    var builder =
-        ClientRegistrations.fromOidcIssuerLocation(providerInfo.getIssuer())
-            .clientId(providerInfo.getClientId())
-            .clientSecret(providerInfo.getClientSecret())
-            .issuerUri(providerInfo.getIssuer());
+  private ClientRegistration buildClientRegistration(
```
Ghost-in-a-Jar: I think it is probably worth adding a test case for this

Ghost-in-a-Jar (changes requested):

service/src/main/java/bio/terra/externalcreds/services/ProviderClientCache.java
samanehsan: `ClientRegistrations.fromOidcIssuerLocation(providerInfo.getIssuer())` makes a GET request to `https://{issuerUri}/.well-known/openid-configuration` so I'm trying to mock this part but still test the rest of the client registration 🤔

common/openapi.yml
```diff
@@ -4,6 +4,24 @@ info:
   description: A manager for external credentials
   version: 0.0.1
 paths:
+  /api/oauth/v1/{provider}/authorization-url:
+    parameters:
+      - $ref: '#/components/parameters/providerParam'
+      - $ref: '#/components/parameters/redirectUriParam'
```
tlangs, line 10: I think we need to do a bit more describing on the parameters here. The client generation needs to know if its a parameter in the path, query, etc. Take a look at https://swagger.io/docs/specification/describing-parameters/.

Looks like it wasn't set up correctly in the first place, so lets take this opportunity to make sure the new endpoints are properly described.

common/openapi.yml
```diff
@@ -4,6 +4,24 @@ info:
...
+    get:
+      summary: Builds an OAuth authorization URL that a user must use to initiate the OAuth dance.
+      tags: [ oauth ]
+      operationId: getAuthUrl
```
tlangs: Can we make this `getAuthorizationUrl`?

service/src/main/java/bio/terra/externalcreds/controllers/OauthApiController.java
```diff
@@ -0,0 +1,37 @@
...
+@Controller
+public record OauthApiController(
+    HttpServletRequest request,
+    ObjectMapper mapper,
+    ProviderService providerService,
+    ExternalCredsSamUserFactory samUserFactory)
+    implements OauthApi {
+
+  @Override
+  public ResponseEntity<String> getAuthUrl(Provider providerName, String redirectUri) {
```
tlangs: This method name will have to change if the `operationId` in the swagger definition changes

service/src/main/java/bio/terra/externalcreds/controllers/OauthApiController.java
```diff
@@ -0,0 +1,37 @@
...
+    var samUser = samUserFactory.from(request);
+
+    var authorizationUrl =
+        providerService.getProviderAuthorizationUrl(
+            samUser.getSubjectId(), providerName.toString(), redirectUri);
+
+    return ResponseEntity.of(authorizationUrl.map(this::jsonString));
```
tlangs: I'd like to take this opportunity to move away from returning JSON Strings. Since there's no structure to this response, encoding it as a JSON String buys us nothing. `text/plain` would be fine.

common/openapi.yml
```diff
@@ -4,6 +4,24 @@ info:
...
+      description: First step to link a new provider. Creates a 1 time use link. There can only be 1 active link per user per provider. The last link created will invalidate any prior links.
+      responses:
+        '200':
+          description: The authorization URL
+          content:
+            application/json:
```
tlangs: Can we make this `text/plain`?

service/src/main/java/bio/terra/externalcreds/config/ProviderPropertiesInterface.java
```diff
@@ -28,6 +28,8 @@ public interface ProviderPropertiesInterface {
 
   String getExternalIdClaim();
 
+  String getUserNameAttributeName();
```
tlangs, line 31: what is this used for?

tlangs (changes requested):

service/src/main/java/bio/terra/externalcreds/config/ProviderPropertiesInterface.java
samanehsan: This gets used when calling the POST oauthcode endpoint (I tested the authorization-url and account linking together to see what I needed to include in the github ClientRegistration). After the token exchange is successful, ECM calls the provider's user info endpoint and [this call](https://github.com/DataBiosphere/terra-external-credentials-manager/blob/dev/service/src/main/java/bio/terra/externalcreds/services/OAuth2Service.java#L137) was failing because that attribute wasn't defined: https://github.com/spring-projects/spring-security/blob/6.1.6/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/userinfo/DefaultOAuth2UserService.java#L103

samanehsan: Also, the field I chose to be the `userNameAttributeName` is "login" which is the github username. This will be stored as the user's `externalUserId` in the linked account table.

common/openapi.yml
samanehsan: Oh interesting! Thanks for catching that

samanehsan: Actually it looks right to me 🤔 The additional info you mentioned is defined here:
- https://github.com/DataBiosphere/terra-external-credentials-manager/blob/091a8fc03204f4bd6b12893ffb792226143b7808/common/openapi.yml#L268
- https://github.com/DataBiosphere/terra-external-credentials-manager/blob/091a8fc03204f4bd6b12893ffb792226143b7808/common/openapi.yml#L274

service/src/main/java/bio/terra/externalcreds/config/ProviderPropertiesInterface.java
samanehsan: RAS has a lot of these values set by default (here is its `userNameAttributeName`): https://github.com/spring-projects/spring-security/blob/6.1.6/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/registration/ClientRegistrations.java#L254

Shakespeared (approved):

common/openapi.yml
tlangs: Oh well would you look at that!

tlangs (approved):

PR #172, opened 2024-02-22
ID-772 Don't Reveal Tomcat Server Version on Error

tlangs:
Jira:  https://broadworkbench.atlassian.net/browse/ID-772

Stop ECM from leaking Tomcat server version when an error is encountered. Usually, this is handled by Spring `ExceptionHandler`s, but a user can make the swagger page serve up a raw Tomcat error. Configuring the error response for Tomcat remediates the issue.


samanehsan (approved):
Thanks for fixing this!

PR #171, opened 2024-02-20
[ID-1050] Add Github account linking

samanehsan:
Jira: https://broadworkbench.atlassian.net/browse/ID-1050

**What:**
Add a new POST `/api/oauth/v1/{provider}/oauthcode` endpoint (supporting both Github and RAS) and deprecate the existing `/api/oidc/v1/{provider}/oauthcode` endpoint

**Why:**
After a user authorizes Github via the Terra UI, Terra will call this endpoint in ECM to exchange the oauth code for a refresh token and store a record for this user's linked account. 

**How:**
When the POST endpoint is called with Github as the provider, call `tokenProviderService.createLink` to exchange the oauthcode for a refresh token, get info about the user from github (this gets the externalUserId), and store an entry in the linked account table.

**Testing:**
1. Run ECM locally and authenticate as a Terra dev user
2. Call `/api/oauth/v1/github/authorization-url` with the callback url "https://bvdp-saturn-dev.appspot.com/oauth_callback"
3. Navigate to the url returned by that endpoint and authorize the Terra github app
4. When you get redirected to Terra, note the "code" and "state" query parameters from the url
5. Call the POST `/api/oauth/v1/github/oauthcode` endpoint with this state and code (this should return a success code and you can confirm by looking at your local ECM database 🎉) 
6. Call the DELETE `/api/oidc/v1/github` endpoint (can also confirm by looking at your local ECM database) - this endpoint required no changes!

PR #173, opened 2024-02-26
[ID-1053] Add API endpoint to return user's access token for a provider (github) 

Shakespeared:
Jira: [ID-1053](https://broadworkbench.atlassian.net/browse/ID-1053) this also covers [ID-1052](https://broadworkbench.atlassian.net/browse/ID-1052)

What:

This adds an API endpoint which returns a user's access token to a particular provider (at this point Github) when called with the userId and providerName as params.

Why:

Because we're adding github account linking, we need to be able to get a user's github access token to authenticate. 

Notes:

I didn’t add tests because all the functions I called are already tested elsewhere.

Manually tested this by linking my GitHub account (I wasn’t able to link directly from my local, but was able to link on dev and then manually copy the refresh token to my local db). Linked by following the instructions in this PR: https://github.com/DataBiosphere/terra-external-credentials-manager/pull/171

Manually verified that I could call the accessToken endpoint with my userId, it returns an access token, and updates my refresh token in the db.  

We could probably do a refactor to extract some helper functions from this if we wanted to streamline PassportProviderService.getRefreshedPassportsAndVisas.
  

[ID-1053]: https://broadworkbench.atlassian.net/browse/ID-1053?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ID-1052]: https://broadworkbench.atlassian.net/browse/ID-1052?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/common/openapi.yml
+++ b/common/openapi.yml
@@ -47,17 +47,12 @@ paths:
           $ref: '#/components/responses/ServerError'
   /api/oauth/v1/{provider}/accesstoken:
     parameters:
-      - $ref: '#/components/parameters/providerParam'
       - $ref: '#/components/parameters/userIdParam'
-      - name: accesstoken
-        description: access token returned by the identity provider
-        in: query
-        schema:
-          type: string
-    post:
+      - $ref: '#/components/parameters/providerParam'
+    get:
       summary: Return an access token from the provider.
       tags: [ oauth ]
-      operationId: accessToken
+      operationId: getProviderAccessToken
       description: Returns a String containing the user's Access Token from the provider
       responses:
         '200':
--- a/common/openapi.yml
+++ b/common/openapi.yml
@@ -45,28 +45,6 @@ paths:
           $ref: '#/components/responses/NotFound'
         '500':
           $ref: '#/components/responses/ServerError'
-  /api/oauth/v1/{provider}/accesstoken:
-    parameters:
-      - $ref: '#/components/parameters/userIdParam'
-      - $ref: '#/components/parameters/providerParam'
-    get:
-      summary: Return an access token from the provider.
-      tags: [ oauth ]
-      operationId: getProviderAccessToken
-      description: Returns a String containing the user's Access Token from the provider
-      responses:
-        '200':
-          description: The user's access token for the requested provider
-          content:
-            text/plain:
-              schema:
-                type: string
-        '400':
-          $ref: '#/components/responses/BadRequest'
-        '404':
-          $ref: '#/components/responses/NotFound'
-        '500':
-          $ref: '#/components/responses/ServerError'
 
   /api/oidc/v1/providers:
     get:
@@ -109,6 +87,29 @@ paths:
           description: Deletes the refresh token and revokes it with the provider.
         '404':
           $ref: '#/components/responses/LinkNotFound'
+        '500':
+          $ref: '#/components/responses/ServerError'
+
+  /api/oidc/v1/{provider}/accessToken:
+    parameters:
+      - $ref: '#/components/parameters/userIdParam'
+      - $ref: '#/components/parameters/providerParam'
+    get:
+      summary: Return an access token from the provider.
+      tags: [ oidc ]
+      operationId: getProviderAccessToken
+      description: Returns a String containing the user's Access Token from the provider
+      responses:
+        '200':
+          description: The user's access token for the requested provider
+          content:
+            text/plain:
+              schema:
+                type: string
+        '400':
+          $ref: '#/components/responses/BadRequest'
+        '404':
+          $ref: '#/components/responses/NotFound'
         '500':
           $ref: '#/components/responses/ServerError'
 
--- a/common/openapi.yml
+++ b/common/openapi.yml
@@ -92,7 +92,6 @@ paths:
 
   /api/oidc/v1/{provider}/access-token:
     parameters:
-      - $ref: '#/components/parameters/userIdParam'
       - $ref: '#/components/parameters/providerParam'
     get:
       summary: Return an access token from the provider.
--- a/common/openapi.yml
+++ b/common/openapi.yml
@@ -45,6 +45,33 @@ paths:
           $ref: '#/components/responses/NotFound'
         '500':
           $ref: '#/components/responses/ServerError'
+  /api/oauth/v1/{provider}/accesstoken:
+    parameters:
+      - $ref: '#/components/parameters/providerParam'
+      - $ref: '#/components/parameters/userIdParam'
+      - name: accesstoken
+        description: access token returned by the identity provider
+        in: query
+        schema:
+          type: string
+    post:
+      summary: Return an access token from the provider.
+      tags: [ oauth ]
+      operationId: accessToken
+      description: Returns a String containing the user's Access Token from the provider
+      responses:
+        '200':
+          description: The user's access token for the requested provider
+          content:
+            text/plain:
+              schema:
+                type: string
+        '400':
+          $ref: '#/components/responses/BadRequest'
+        '404':
+          $ref: '#/components/responses/NotFound'
+        '500':
+          $ref: '#/components/responses/ServerError'
 
   /api/oidc/v1/providers:
     get:
@@ -296,6 +323,13 @@ components:
       required: true
       schema:
         $ref: '#/components/schemas/Provider'
+    userIdParam:
+      name: userId
+      description: the Sam User Id for a user
+      in: query
+      required: true
+      schema:
+        type: string
     redirectUriParam:
       name: redirectUri
       description: oidc redirect uri
--- a/common/openapi.yml
+++ b/common/openapi.yml
@@ -46,6 +46,28 @@ paths:
         '500':
           $ref: '#/components/responses/ServerError'
 
+  /api/oidc/v1/{provider}/access-token:
+    parameters:
+      - $ref: '#/components/parameters/providerParam'
+    get:
+      summary: Return an access token from the provider.
+      tags: [ oauth ]
+      operationId: getProviderAccessToken
+      description: Returns a String containing the user's Access Token from the provider
+      responses:
+        '200':
+          description: The user's access token for the requested provider
+          content:
+            text/plain:
+              schema:
+                type: string
+        '400':
+          $ref: '#/components/responses/BadRequest'
+        '404':
+          $ref: '#/components/responses/NotFound'
+        '500':
+          $ref: '#/components/responses/ServerError'
+
   /api/oidc/v1/providers:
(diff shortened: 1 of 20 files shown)


service/src/main/java/bio/terra/externalcreds/services/ProviderClientCache.java
```diff
@@ -60,7 +60,7 @@ public ClientRegistration buildClientRegistration(
                     .toList()
                     .get(0);
             yield ClientRegistration.withRegistrationId(providerName)
-                .authorizationGrantType(AuthorizationGrantType.AUTHORIZATION_CODE)
```
tlangs: why did this change?

service/src/main/java/bio/terra/externalcreds/services/TokenProviderService.java
```diff
@@ -79,4 +83,51 @@ public void logLinkCreation(
...
+          String.format("No refresh token found for provider %s", providerName));
+    }
+
+    // exchange refresh token for access token
+    var accessTokenResponse =
+        oAuth2Service.authorizeWithRefreshToken(
+            clientRegistration, new OAuth2RefreshToken(linkedAccount.getRefreshToken(), null));
+
+    // save the linked account with the new refresh token to replace the old one
+    var linkedAccountWithRefreshToken =
```
tlangs: `linkedAccountWithRefreshToken` is never used, so it doesn't need to be saved as a `var`

tlangs (changes requested):

service/src/main/java/bio/terra/externalcreds/services/ProviderClientCache.java
Shakespeared: In the [github docs for refreshing a user access](https://docs.github.com/en/apps/creating-github-apps/authenticating-with-a-github-app/refreshing-user-access-tokens#refreshing-a-user-access-token-with-a-refresh-token) token it says the `grant_type` must be `refresh_token` to exchange the refresh for the access token

tlangs: And this doesn't impact the oauth login process where the Authorization Code is needed to complete the flow?

Shakespeared: ah I see, you're right looks like we need to have both at different times

Shakespeared: what do you think about building a new client registration where I'm getting the refresh token vs adding complexity to the cache?

tlangs: The cache is going to be very important to the performance of ECM. I think `TokenServiceProvider` might need its own cache.

tlangs: Looks like a `TokenClientCache`, analogous to `ProviderClientCache`, is what we want. Maybe renaming `ProviderClientCache` to `ProviderOAuthClientCache`, and creating a new `ProviderTokenClientCache`, is the way to go.

Shakespeared: Yeah that makes sense, this does seem like something we want cached. I can take a quick stab at it!

common/openapi.yml
```diff
@@ -45,6 +45,28 @@ paths:
           $ref: '#/components/responses/NotFound'
         '500':
           $ref: '#/components/responses/ServerError'
+  /api/oauth/v1/{provider}/accesstoken:
```
tlangs: I _think_ we want this to be under `/api/oidc/` since it's not part of the OAuth flow. Also, can we camelCase `accessToken`?

service/src/main/java/bio/terra/externalcreds/services/ProviderClientCache.java
```diff
@@ -31,15 +31,15 @@ public ProviderClientCache(ExternalCredsConfig externalCredsConfig) {
     this.externalCredsConfig = externalCredsConfig;
   }
 
-  @Cacheable(cacheNames = "providerClients", sync = true)
+  @Cacheable(cacheNames = "providerTokenClients", sync = true)
```
tlangs: I think this one should be names `providerOAuthClients` to distinguish it from the token clients

tlangs: Could we also rename the class to `ProviderOAuthClientCache`

tlangs (changes requested):

common/openapi.yml
Shakespeared: yeah I wasn't sure whether we wanted it under oidc or oauth, I can change that 👍

Shakespeared: if we're camelcasing accessToken, would we also want to camelcase oauthcode and authorization-url for consistency?

common/openapi.yml
```diff
@@ -90,6 +90,29 @@ paths:
         '500':
           $ref: '#/components/responses/ServerError'
 
+  /api/oidc/v1/{provider}/accessToken:
```
tlangs: ack, it actually looks like ecm kebab-cases its URLs! Lets do that to be consistent... Sorry!

Shakespeared: In that case I think we should also update oauthcode to be oauth-code (the UI is the only thing calling that right?)

common/openapi.yml
```diff
@@ -90,6 +90,29 @@ paths:
         '500':
           $ref: '#/components/responses/ServerError'
 
+  /api/oidc/v1/{provider}/access-token:
+    parameters:
+      - $ref: '#/components/parameters/userIdParam'
```
tlangs: We won't need the userId as a param, as we'll be able to resolve that from the request headers. Looks the controllers use `var samUser = samUserFactory.from(request);` to get the Sam user.

service/src/main/java/bio/terra/externalcreds/services/ProviderService.java
```diff
@@ -39,7 +39,8 @@
 @Slf4j
 public class ProviderService {
   public final ExternalCredsConfig externalCredsConfig;
-  public final ProviderClientCache providerClientCache;
+  public final ProviderOAuthClientCache providerOAuthClientCache;
+  public final ProviderTokenClientCache providerTokenClientCache;
```
tlangs, line 43: It doesn't look like `ProviderService` uses the `providerTokenClientCache`, so we can just have it be a member if the `TokenProviderService` subclass

service/src/main/java/bio/terra/externalcreds/services/ProviderTokenClientCache.java
```diff
@@ -0,0 +1,81 @@
...
+@Component
+@Slf4j
+public class ProviderTokenClientCache {
+  private final ExternalCredsConfig externalCredsConfig;
+
+  public ProviderTokenClientCache(ExternalCredsConfig externalCredsConfig) {
+    this.externalCredsConfig = externalCredsConfig;
+  }
+
+  @Cacheable(cacheNames = "providerOAuthClients", sync = true)
```
tlangs: Should be named `providerTokenClients`

service/src/main/java/bio/terra/externalcreds/services/ProviderTokenClientCache.java
```diff
@@ -0,0 +1,81 @@
...
+  public Optional<ClientRegistration> getProviderClient(String providerName) {
+    log.info("Loading ProviderClient {}", providerName);
+    return Optional.ofNullable(externalCredsConfig.getProviders().get(providerName))
+        .map(p -> buildClientRegistration(providerName, p));
+  }
+
+  @Scheduled(fixedRateString = "6", timeUnit = TimeUnit.HOURS)
+  @CacheEvict(allEntries = true, cacheNames = "providerOAuthClients")
```
tlangs: Same here, should be named `providerTokenClients`

service/src/main/java/bio/terra/externalcreds/services/ProviderTokenClientCache.java
```diff
@@ -0,0 +1,81 @@
...
+              .clientSecret(providerInfo.getClientSecret())
+              .issuerUri(providerInfo.getIssuer());
+          case GITHUB -> {
+            String redirectUri =
+                providerInfo.getAllowedRedirectUriPatterns().stream()
+                    .map(Pattern::toString)
+                    .toList()
+                    .get(0);
+            yield ClientRegistration.withRegistrationId(providerName)
+                .authorizationGrantType(AuthorizationGrantType.AUTHORIZATION_CODE)
```
tlangs: Here's where we'll change `AUTHORIZATION_CODE` to `REFRESH_TOKEN`

tlangs (changes requested):

service/src/main/java/bio/terra/externalcreds/services/ProviderService.java
Shakespeared: If I remove it from ProviderService I get a `cannot find symbol` error

service/src/main/java/bio/terra/externalcreds/services/ProviderTokenClientCache.java
Shakespeared: huh I could have sworn I changed that

common/openapi.yml
```diff
@@ -90,6 +90,28 @@ paths:
         '500':
           $ref: '#/components/responses/ServerError'
 
+  /api/oidc/v1/{provider}/access-token:
```
samanehsan: I think we should add this as an `api/oauth/v1` endpoint instead since Github is not an oidc provider

Shakespeared: that's how I had it originally, but Trevyn was arguing that it belongs more under `oidc` because it's not really part of the oauth flow (i think that was the reasoning, it was in an oudated comment)

common/openapi.yml
tlangs: I think we can just leave that one, since we want to get this out in front of Workflows ASAP. We can come back to it.

Shakespeared: yeah agreed best not to do it now, let's just track that as one of our follow-up items

tlangs (approved):

samanehsan (approved):

PR #174, opened 2024-02-28
[ID-1117] Add fence_account_key and distributed_lock tables to db

samanehsan:
**Jira:** https://broadworkbench.atlassian.net/browse/ID-1117

**What:**
- Add a new `fence_account_key` table that is equivalent to [the one in Bond](https://github.com/DataBiosphere/bond/blob/9151bc752b8e1e3ce5f646b8d177278ccbad1582/bond_app/fence_token_storage.py#L22) and a `distributed_lock` table to store the lock information about fence tokens (or eventually passports).

**Why:**
The Identiteam is moving Bond's functionality to authenticate with fence account providers into ECM, so ECM needs to store information about fence account keys.

**How:**
- Add a new changeset to the liquibase change log.
- Added corresponding model and DAO classes

**Testing:**
I ran the ECM server locally and confirmed that the new table was created.


service/src/main/java/bio/terra/externalcreds/models/DistributedLock.java
```diff
@@ -0,0 +1,15 @@
+package bio.terra.externalcreds.models;
+
+import java.sql.Timestamp;
+import org.immutables.value.Value;
+
+@Value.Immutable
+public interface DistributedLock {
```
tlangs:
```suggestion
public interface DistributedLock extends WithDistributedLock {
```

This, along with `./gradlew assemble`, will let you uncomment the builder.

service/src/main/java/bio/terra/externalcreds/models/FenceAccountKey.java
```diff
@@ -0,0 +1,16 @@
+package bio.terra.externalcreds.models;
+
+import java.sql.Timestamp;
+import java.util.Optional;
+import org.immutables.value.Value;
+
+@Value.Immutable
+public interface FenceAccountKey {
```
tlangs: Same as above

service/src/main/java/bio/terra/externalcreds/models/DistributedLock.java
samanehsan: Oh! I had to add `extends WithDistributedLock` first and then run `./gradlew assemble` in order to generate that interface

samanehsan: Hmm I'm still getting an error about the builder being final 🤔

samanehsan: Oh `./gradlew assemble` fixed the error!

service/src/main/resources/db/changelog/changesets/20240228_add_fence_account_key_and_dist_lock_table.yaml
```diff
@@ -0,0 +1,45 @@
+databaseChangeLog:
+  - changeSet:
+      id: "add_fence_account_key_and_dist_lock_table"
+      author: sehsan
+      changes:
+        - createTable:
+            tableName: fence_account_key
```
samanehsan: I think we also need a foreign key to the user's `linked_account` record, so that we can retrieve a user's fence account key by user_id and provider (similar to [ga4gh_passport](https://github.com/DataBiosphere/terra-external-credentials-manager/blob/dev/service/src/main/resources/db/changelog/changesets/20210521_initial_schema.yaml#L56)). Does that make sense @tlangs?

tlangs: That does make sense!

service/src/main/java/bio/terra/externalcreds/models/FenceAccountKey.java
```diff
@@ -0,0 +1,18 @@
+package bio.terra.externalcreds.models;
+
+import java.sql.Timestamp;
+import java.util.Optional;
+import org.immutables.value.Value;
+
+@Value.Immutable
+public interface FenceAccountKey extends WithFenceAccountKey {
+  Optional<Integer> getId();
```
samanehsan, line 9: FYI this is optional because when we pass a `FenceAccountKey` object to `insertFenceAccountKey`, it won't have an id at that point.

service/src/main/resources/db/changelog/changesets/20240228_add_fence_account_key_and_dist_lock_table.yaml
```diff
@@ -0,0 +1,52 @@
...
+                    nullable: false
+              - column:
+                  name: linked_account_id
+                  type: int
+                  constraints:
+                    nullable: false
+                    references: linked_account(id)
+                    foreignKeyName: fk_linked_account_id
+              - column:
+                  name: key_json
```
Ghost-in-a-Jar: nit: maybe this should be called key_json_base64?

service/src/main/java/bio/terra/externalcreds/dataAccess/DistributedLockDAO.java
```diff
@@ -0,0 +1,80 @@
...
+
+  /**
+   * @param distributedLock The DistributedLock to insert
+   * @return distributedLock The DistributedLock that was inserted
+   */
+  @WithSpan
+  public DistributedLock upsertDistributedLock(DistributedLock distributedLock) {
+    var query =
+        "INSERT INTO distributed_lock (lock_name, user_id, expires_at)"
+            + " VALUES (:lockName, :userId, :expiresAt)";
```
Ghost-in-a-Jar, line 57: I think for this to be an upsert you need an `ON CONFLICT` clause. Otherwise if there is already a lock_name+user_id youll get an error. Or maybe this fn should be called `insertDistributedLock`.

service/src/main/java/bio/terra/externalcreds/dataAccess/DistributedLockDAO.java
```diff
@@ -0,0 +1,80 @@
...
+   * @param lockName The name of the lock, e.g {provider}-createKey
+   * @param userId The Sam user id
+   * @return Optional<DistributedLock> An optional containing the distributed lock with this
+   *     lockName and userId (or empty)
+   */
+  @WithSpan
+  public Optional<DistributedLock> getDistributedLock(String lockName, String userId) {
+    var namedParameters =
+        new MapSqlParameterSource().addValue("lockName", lockName).addValue("userId", userId);
+    var query = "SELECT * FROM distributed_lock WHERE lock_name = :lockName AND user_id = :userId";
```
Ghost-in-a-Jar: This is nitpicky but I think it is best to name each column in a DAO query.

service/src/main/java/bio/terra/externalcreds/dataAccess/FenceAccountKeyDAO.java
```diff
@@ -0,0 +1,87 @@
...
+  public FenceAccountKeyDAO(NamedParameterJdbcTemplate jdbcTemplate) {
+    this.jdbcTemplate = jdbcTemplate;
+  }
+
+  @WithSpan
+  public Optional<FenceAccountKey> getFenceAccountKey(String userId, String providerName) {
+    var namedParameters =
+        new MapSqlParameterSource("userId", userId).addValue("providerName", providerName);
+    var query =
+        "SELECT p.* FROM fence_account_key p"
```
Ghost-in-a-Jar: Same comment as above I would name each column instead of using *.

Ghost-in-a-Jar: also nitpicky but I would use something more descriptive than 'p' for fence_account_key's alias, maybe 'fence' or 'fak'. Just to make the query easier to read.

service/src/main/java/bio/terra/externalcreds/dataAccess/FenceAccountKeyDAO.java
```diff
@@ -0,0 +1,87 @@
...
+            .addValue("linkedAccountId", fenceAccountKey.getLinkedAccountId())
+            .addValue("keyJson", base64Encode(fenceAccountKey.getKeyJson()))
+            .addValue("expiresAt", fenceAccountKey.getExpiresAt());
+
+    // generatedKeyHolder will hold the id returned by the query as specified by the RETURNING
+    // clause
+    var generatedKeyHolder = new GeneratedKeyHolder();
+    jdbcTemplate.update(query, namedParameters, generatedKeyHolder);
+
+    return fenceAccountKey.withId(Objects.requireNonNull(generatedKeyHolder.getKey()).intValue());
```
Ghost-in-a-Jar, line 73: I love that you can get the id back from the insert query, there are limitations on this in scala libs that I have used.

Ghost-in-a-Jar (changes requested):
Looks good overall, just some small changes requested

service/src/main/resources/db/changelog/db.changelog-master.yaml
```diff
@@ -29,3 +29,6 @@ databaseChangeLog:
   - include:
       file: changesets/20220412_add_last_encrypt_timestamp_to_ssh_key_pair_table.yaml
       relativeToChangelogFile: true
+  - include:
```
tlangs, line 32: There's a merge conflict here. Make sure this changeset comes _after_ the just-merged changeset.

service/src/main/java/bio/terra/externalcreds/dataAccess/FenceAccountKeyDAO.java
```diff
@@ -0,0 +1,87 @@
...
+@Repository
+@Slf4j
+public class FenceAccountKeyDAO {
+
+  private static final RowMapper<FenceAccountKey> FENCE_ACCOUNT_KEY_ROW_MAPPER =
+      ((rs, rowNum) ->
+          new FenceAccountKey.Builder()
+              .id(rs.getInt("id"))
+              .linkedAccountId(rs.getInt("linked_account_id"))
+              .keyJson(base64Decode(rs.getString("key_json")))
```
tlangs: So, I looked at the Bond code, and its storing the keys as raw json instead of base64. I thought it about it a little bit more, and storing the keys as JSON or JSONB might actually be better, since we could then query into the keys to find info. That may be useful for debugging or analysis.

tlangs (changes requested):

service/src/main/java/bio/terra/externalcreds/dataAccess/FenceAccountKeyDAO.java
tlangs: Apologies for the whiplash...

service/src/main/resources/db/changelog/changesets/20240228_add_fence_account_key_and_dist_lock_table.yaml
```diff
@@ -0,0 +1,52 @@
...
+            columns:
+              - column:
+                  name: id
+                  type: int
+                  autoIncrement: true
+                  constraints:
+                    primaryKey: true
+                    nullable: false
+              - column:
+                  name: linked_account_id
```
tlangs: Oh! Can we make this cascade on delete? That way, if someone deletes their linked account, the key will automatically get deleted too.

service/src/main/java/bio/terra/externalcreds/dataAccess/FenceAccountKeyDAO.java
samanehsan: No worries! That's a good point about search-ability

service/src/main/resources/db/changelog/changesets/20240228_add_fence_account_key_and_dist_lock_table.yaml
samanehsan: Oh I assumed that was the default behavior 😅 I'll double check!

service/src/main/java/bio/terra/externalcreds/dataAccess/FenceAccountKeyDAO.java
```diff
@@ -0,0 +1,81 @@
...
+            + " INNER JOIN linked_account la ON la.id = fence.linked_account_id"
+            + " WHERE la.user_id = :userId"
+            + " AND la.provider = :provider::provider_enum";
+    return Optional.ofNullable(
+        DataAccessUtils.singleResult(
+            jdbcTemplate.query(query, namedParameters, FENCE_ACCOUNT_KEY_ROW_MAPPER)));
+  }
+
+  @WithSpan
+  public FenceAccountKey insertFenceAccountKey(FenceAccountKey fenceAccountKey) {
```
tlangs: Can we rename this `upsertFenceAccountKey` and do the same `ON CONFLICT` logic?

Ghost-in-a-Jar (approved):

service/src/main/java/bio/terra/externalcreds/dataAccess/FenceAccountKeyDAO.java
Ghost-in-a-Jar: im confused i thought it was storing the keys as b64?

service/src/main/java/bio/terra/externalcreds/models/DistributedLock.java
```diff
@@ -0,0 +1,15 @@
...
+import java.sql.Timestamp;
+import org.immutables.value.Value;
+
+@Value.Immutable
+public interface DistributedLock extends WithDistributedLock {
+  String getLockName();
+
+  String getUserId();
+
+  Timestamp getExpiresAt();
```
tlangs: We should probably use `java.time.Instant` here instead of the SQL Timestamp.

service/src/main/java/bio/terra/externalcreds/dataAccess/DistributedLockDAO.java
tlangs: Actually, I think we want to error here. If a lock already exists, ECM should not proceed with whatever action it's trying to do.

samanehsan: Yeah, that's what I was thinking

service/src/main/java/bio/terra/externalcreds/dataAccess/FenceAccountKeyDAO.java
samanehsan: Is it just when there is a conflict on the `id` and `linked_account_id`?

samanehsan: Ah I had to add a unique constraint on the `linked_account_id` in order to use is as the conflict criteria

tlangs (approved):

service/src/main/java/bio/terra/externalcreds/dataAccess/FenceAccountKeyDAO.java
tlangs: Bond stores them as plain text.

PR #176, opened 2024-03-04
ID-1131 ECM Code Cleanup

tlangs:
Jira: https://broadworkbench.atlassian.net/browse/ID-1131

While we were developing GitHub Account Linking in ECM, we made some decisions to defer maintenance to get the feature out the door. Now, we're going back through and making improvements!

The biggest change is migrating Providers to use an `enum` instead of being represented as a `String`. This brings type safety to provider types, simplifying the code in a lot of places.

### DIFF
--- a/.run/ExternalCredsApplication.run.xml
+++ b/.run/ExternalCredsApplication.run.xml
@@ -1,11 +1,6 @@
 <component name="ProjectRunConfigurationManager">
   <configuration default="false" name="ExternalCredsApplication" type="SpringBootApplicationConfigurationType" factoryName="Spring Boot">
-    <module name="externalcreds.service.main" />
-    <option name="SPRING_BOOT_MAIN_CLASS" value="bio.terra.externalcreds.ExternalCredsWebApplication" />
-    <option name="HIDE_BANNER" value="true" />
     <option name="ACTIVE_PROFILES" value="human-readable-logging" />
-    <option name="ALTERNATIVE_JRE_PATH" />
-    <option name="SHORTEN_COMMAND_LINE" value="NONE" />
     <additionalParameters>
       <param>
         <option name="enabled" value="true" />
@@ -15,7 +10,21 @@
     </additionalParameters>
     <envs>
       <env name="GOOGLE_APPLICATION_CREDENTIALS" value="service/src/main/resources/rendered/ecm-sa.json" />
+      <env name="ECM_LOG_APPENDER" value="Console-Standard" />
     </envs>
+    <option name="HIDE_BANNER" value="true" />
+    <module name="externalcreds.service.main" />
+    <option name="SPRING_BOOT_MAIN_CLASS" value="bio.terra.externalcreds.ExternalCredsWebApplication" />
+    <extension name="net.ashald.envfile">
+      <option name="IS_ENABLED" value="false" />
+      <option name="IS_SUBST" value="false" />
+      <option name="IS_PATH_MACRO_SUPPORTED" value="false" />
+      <option name="IS_IGNORE_MISSING_FILES" value="false" />
+      <option name="IS_ENABLE_EXPERIMENTAL_INTEGRATIONS" value="false" />
+      <ENTRIES>
+        <ENTRY IS_ENABLED="true" PARSER="runconfig" IS_EXECUTABLE="false" />
+      </ENTRIES>
+    </extension>
     <method v="2">
       <option name="Make" enabled="true" />
     </method>
--- a/service/src/main/java/bio/terra/externalcreds/controllers/OidcApiController.java
+++ b/service/src/main/java/bio/terra/externalcreds/controllers/OidcApiController.java
@@ -114,7 +114,7 @@ public record OidcApiController(
   @Override
   public ResponseEntity<String> getProviderPassport(PassportProvider passportProvider) {
     var samUser = samUserFactory.from(request);
-    var provider = Provider.fromValue(passportProvider.toString());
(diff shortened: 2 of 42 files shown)


service/src/main/java/bio/terra/externalcreds/controllers/OidcApiController.java
```diff
@@ -50,47 +50,39 @@ public ResponseEntity<List<String>> listProviders() {
...
     var samUser = samUserFactory.from(request);
-    var linkedAccount =
-        linkedAccountService.getLinkedAccount(samUser.getSubjectId(), providerName.toString());
+    var linkedAccount = linkedAccountService.getLinkedAccount(samUser.getSubjectId(), provider);
     return ResponseEntity.of(linkedAccount.map(OpenApiConverters.Output::convert));
   }
 
   @Override
-  public ResponseEntity<String> getAuthUrl(Provider providerName, String redirectUri) {
+  public ResponseEntity<String> getAuthUrl(Provider provider, String redirectUri) {
```
samanehsan, line 53: Is this route still being used in the Terra UI/other services? (Same with the createLink route) It would be nice to delete that either in this PR or a follow-on! 🔪

tlangs: Yeah, its still using this. I can fold that work into this ticket, I think.

samanehsan: Hmm if it's still being used, I don't think removing those routes needs to block this PR from being merged!

service/src/main/java/bio/terra/externalcreds/services/PassportProviderService.java
```diff
@@ -206,14 +202,7 @@ void authAndRefreshPassport(LinkedAccount linkedAccount) {
...
       LinkedAccount linkedAccount) {
     var clientRegistration =
-        providerOAuthClientCache
-            .getProviderClient(linkedAccount.getProviderName())
-            .orElseThrow(
-                () ->
-                    new ExternalCredsException(
-                        String.format(
-                            "Unable to find configs for the provider: %s",
-                            linkedAccount.getProviderName())));
```
Ghost-in-a-Jar, line 216: Did this error get moved elsewhere?

Ghost-in-a-Jar (approved):

service/src/main/java/bio/terra/externalcreds/services/TokenProviderService.java
```diff
@@ -39,39 +38,34 @@ public TokenProviderService(
...
+                  oAuth2State.getRedirectUri(),
+                  new HashSet<>(providerInfo.getScopes()),
+                  encodedState,
+                  providerClient)
+              .getLeft();
+      var linkedAccount = linkedAccountService.upsertLinkedAccount(account);
+      logLinkCreation(Optional.of(linkedAccount), auditLogEventBuilder);
+      return linkedAccount;
+    } catch (OAuth2AuthorizationException oauthEx) {
+      logLinkCreation(Optional.empty(), auditLogEventBuilder);
```
samanehsan, line 66: Thanks for catching/fixing the logging issue here!

samanehsan (approved):
Thanks for doing all of this code clean-up! :octocat:

service/src/main/java/bio/terra/externalcreds/services/PassportProviderService.java
tlangs: Yup! Since `provider` is now a type-safe enum, we only have to error if we don't have the key in the map of provider configs. So, I made a convenience method to check for key existence in the map, and migrated all calls to use that.

https://github.com/DataBiosphere/terra-external-credentials-manager/pull/176/files#diff-9f53bef3d9981d655b2d7d6ca430ab4fc517cd4e76c2e13b19d03ce51ec10e81R19-R25

Ghost-in-a-Jar (approved):

service/src/main/java/bio/terra/externalcreds/controllers/OidcApiController.java
```diff
@@ -43,71 +41,28 @@ public record OidcApiController(
 
   @Override
   public ResponseEntity<List<String>> listProviders() {
-    var providerNames = new ArrayList<>(providerService.getProviderList());
-    Collections.sort(providerNames);
-
-    return ResponseEntity.ok(providerNames);
+    return oauthApiController.listProviders();
```
samanehsan, line 44: This is so much nicer than duplicating the code!

samanehsan (approved):
The latest changes look good to me!

.run/ExternalCredsApplication.run.xml
```diff
@@ -1,11 +1,6 @@
 <component name="ProjectRunConfigurationManager">
   <configuration default="false" name="ExternalCredsApplication" type="SpringBootApplicationConfigurationType" factoryName="Spring Boot">
-    <module name="externalcreds.service.main" />
-    <option name="SPRING_BOOT_MAIN_CLASS" value="bio.terra.externalcreds.ExternalCredsWebApplication" />
-    <option name="HIDE_BANNER" value="true" />
     <option name="ACTIVE_PROFILES" value="human-readable-logging" />
-    <option name="ALTERNATIVE_JRE_PATH" />
-    <option name="SHORTEN_COMMAND_LINE" value="NONE" />
```
samanehsan, line 8: Just curious, why were these two options removed?

tlangs: Honestly, idk why this is even checked in...

PR #178, opened 2024-03-06
ID-1125 ECM Migrates Data from Bond

tlangs:
Jira: https://broadworkbench.atlassian.net/browse/ID-1125

We want to retire Bond and put its functionality into ECM. This presents a problem for users, since if they link an account in Bond, and then we cut over to ECM, they're linked account will go away! To resolve this, we can make ECM read from Bond's Datastore and dynamically update its own database if it finds any info. 

This is temporary logic that will exist for a maximum of 30 days, at which point all fence account links will have expired.


service/src/main/java/bio/terra/externalcreds/dataAccess/FenceAccountKeyDAO.java
```diff
@@ -32,6 +33,23 @@ public FenceAccountKeyDAO(NamedParameterJdbcTemplate jdbcTemplate) {
     this.jdbcTemplate = jdbcTemplate;
   }
 
+  public Optional<FenceAccountKey> getFenceAccountKey(LinkedAccount linkedAccount) {
```
samanehsan, line 37: Does this also need a `@WithSpan` annotation? I realized I don't actually know what that does 🤔

service/src/main/java/bio/terra/externalcreds/models/BondFenceServiceAccountEntity.java
```diff
@@ -0,0 +1,23 @@
+package bio.terra.externalcreds.models;
+
+import com.google.cloud.datastore.Key;
+import java.time.Instant;
+import org.immutables.value.Value;
+
+@Value.Immutable
+public interface BondFenceServiceAccountEntity extends WithBondFenceServiceAccountEntity {
```
samanehsan, line 9: This looks very similar to `FenceAccountKey` 🤔  How are they related?

service/src/main/java/bio/terra/externalcreds/services/FenceKeyRetriever.java
```diff
@@ -0,0 +1,135 @@
...
+
+  @Retryable(
+      retryFor = {DistributedLockException.class},
+      maxAttemptsExpression = "${retry.getFenceAccountKey.maxAttempts}",
+      backoff = @Backoff(delayExpression = "${retry.getFenceAccountKey.delay}"))
+  public Optional<FenceAccountKey> createFenceAccountKey(LinkedAccount linkedAccount) {
+    var maybeKey = fenceAccountKeyService.getFenceAccountKey(linkedAccount);
+    return maybeKey.or(
+        () -> {
+          var newLock =
```
samanehsan: I think this would be easier to read if this lock logic was extracted into a separate function 🔒

service/src/main/java/bio/terra/externalcreds/services/FenceKeyRetriever.java
```diff
@@ -0,0 +1,135 @@
...
+    WebClient.ResponseSpec response =
+        WebClient.create(keyEndpoint.get())
+            .post()
+            .header("Authorization", "Bearer " + accessToken.getAccessToken().getTokenValue())
+            .retrieve();
+    String responseBody =
+        response
+            .onStatus(HttpStatusCode::isError, clientResponse -> Mono.empty())
+            .bodyToMono(String.class)
+            .block(Duration.of(11, ChronoUnit.SECONDS));
```
samanehsan, line 146: Maybe this should be a configurable value

service/src/test/java/bio/terra/externalcreds/services/ProviderTokenClientCacheTest.java
```diff
@@ -48,4 +43,35 @@ void testGitHubBuildClientRegistration() {
     assertEquals(providerInfo.getClientSecret(), gitHubClient.getClientSecret());
     assertEquals(redirectUri, gitHubClient.getRedirectUri());
   }
+
+  @Test
+  void testRasBuildClientRegistration() {
```
samanehsan, line 47: Thanks for adding this test! 💯

service/src/main/java/bio/terra/externalcreds/services/FenceProviderService.java
```diff
@@ -0,0 +1,135 @@
...
+    var ecmLinkedAccount = linkedAccountService.getLinkedAccount(userId, provider);
+    var bondLinkedAccount = bondService.getLinkedAccount(userId, provider);
+    if (ecmLinkedAccount.isPresent() && bondLinkedAccount.isPresent()) {
+      var ecmExpires = ecmLinkedAccount.get().getExpires();
+      var bondExpires = bondLinkedAccount.get().getExpires();
+      if (ecmExpires.after(bondExpires) || ecmExpires.equals(bondExpires)) {
+        // ECM is up to date
+        return ecmLinkedAccount;
+      } else {
+        // ECM is out of date. Update it with the Bond data and return the new ECM Linked Account
```
samanehsan, line 61: When would the ECM linked account have an older expiration time than the one in Bond (since it would have been ported from the Bond db)?

service/src/test/java/bio/terra/externalcreds/BaseTest.java
```diff
@@ -14,6 +15,7 @@
 @ActiveProfiles({"test", "human-readable-logging"})
 @Transactional
 @Rollback
+@DirtiesContext(classMode = DirtiesContext.ClassMode.AFTER_CLASS)
```
samanehsan, line 18: What does this annotation do?

service/src/main/java/bio/terra/externalcreds/models/BondFenceServiceAccountEntity.java
samanehsan: Oh it looks like `BondFenceServiceAccountEntity` is the model for the info we get from Bond and `FenceAccountKey` is just used for the ECM representation?

service/src/main/java/bio/terra/externalcreds/services/FenceProviderService.java
tlangs: If, somehow, someone connected to Bond directly to renew their link. Probably won't happen, but still a possibility.

service/src/main/java/bio/terra/externalcreds/models/BondFenceServiceAccountEntity.java
tlangs: Exactly! The `BondDatastoreDAO` is bond-specific, and the `BondService` handles converting Bond models to ECM ones. I didn't want to encode a lot of business logic in the DAO layer.

service/src/main/java/bio/terra/externalcreds/dataAccess/FenceAccountKeyDAO.java
tlangs: I believe it does

tlangs: `@WithSpan` allows us to trace the performance of the DB query separate from the API request that triggered it.

service/src/test/java/bio/terra/externalcreds/BaseTest.java
tlangs: It makes sure that Spring throws out its state after each test file. Without this, Spring keeps around its db connections for completed tests, and we end up running out of db connections!

service/src/main/java/bio/terra/externalcreds/services/FenceKeyRetriever.java
tlangs: Probably. I'll have to wait until the provider config is merged to do that though.

service/src/main/java/bio/terra/externalcreds/services/FenceKeyRetriever.java
```diff
@@ -0,0 +1,146 @@
...
+  private Optional<FenceAccountKey> retrieveNewKeyFromFence(LinkedAccount linkedAccount) {
+    var newLock =
+        new DistributedLock.Builder()
+            .lockName("createFenceKey-" + linkedAccount.getProvider())
+            .userId(linkedAccount.getUserId())
+            .expiresAt(Instant.now().plus(1, ChronoUnit.MINUTES))
+            .build();
+    Optional<DistributedLock> existingLock =
+        distributedLockDAO.getDistributedLock(newLock.getLockName(), linkedAccount.getUserId());
+    if (existingLock.isPresent()) {
```
samanehsan, line 67: If the fence account retrieval/upsert fails it won't clean up the lock. What about deleting the existing lock if it is expired? Or updating `obtainLockAndGetNewKey` so that it deletes the lock even if it fails?

samanehsan: Also thanks for splitting up this function! I find it much easier to follow what it's doing!

service/src/main/java/bio/terra/externalcreds/services/FenceProviderService.java
samanehsan: Oh I see!

service/src/main/java/bio/terra/externalcreds/services/FenceKeyRetriever.java
tlangs: Just made that happen!

service/src/main/java/bio/terra/externalcreds/controllers/OauthApiController.java
```diff
@@ -38,7 +40,13 @@ public ResponseEntity<List<String>> listProviders() {
   @Override
   public ResponseEntity<LinkInfo> getLink(Provider provider) {
     var samUser = samUserFactory.from(request);
-    var linkedAccount = linkedAccountService.getLinkedAccount(samUser.getSubjectId(), provider);
+    var linkedAccount =
+        switch (provider) {
+          case RAS, GITHUB -> linkedAccountService.getLinkedAccount(
+              samUser.getSubjectId(), provider);
+          case FENCE, DCF_FENCE, KIDS_FIRST, ANVIL -> fenceProviderService.getLinkedFenceAccount(
```
Ghost-in-a-Jar, line 47: yay for pattern matching

service/src/main/java/bio/terra/externalcreds/dataAccess/BondDatastoreDAO.java
```diff
@@ -0,0 +1,88 @@
...
+    var entity = datastore.get(key);
+    var bondFenceServiceAccountEntity =
+        new BondFenceServiceAccountEntity.Builder()
+            .key(key)
+            .expiresAt(
+                entity
+                    .getTimestamp(BondFenceServiceAccountEntity.expiresAtName)
+                    .toDate()
+                    .toInstant())
+            .keyJson(entity.getString(BondFenceServiceAccountEntity.keyJsonName))
```
Ghost-in-a-Jar, line 63: does this need to be decoded or anything

Ghost-in-a-Jar: or the key i mean

Ghost-in-a-Jar (approved):
This looks good i think you did a good job of covering test cases. One thing that would be helpful for reviewing such a large pr in the future is to add some more detail about the logic flow to the description and maybe inline pr comments.

samanehsan (approved):
I think this looks good once it's up-to-date with the changes I just merged! :octocat:

tlangs:
Yesss I got coverage!

PR #180, opened 2024-03-11
[ID-1120] Add /api/serviceaccount/v1/{provider}/key endpoint

samanehsan:
**Jira:** https://broadworkbench.atlassian.net/browse/ID-1120

**What:** 
Add an `/api/serviceaccount/v1/{provider}/key` endpoint that uses the requesting user's id and the provider to look up a fenceAccountKey from the database, and return it if it exists and is not expired.

**Why:**
Identiteam is getting rid of Bond and adding its functionality to ECM to consolidate these similar services. This new endpoint is equivalent to the Bond `/api/link/v1/{provider}/serviceaccount/key` [endpoint](https://bond.dsde-dev.broadinstitute.org/api/docs/#/).

### DIFF
--- a/common/openapi.yml
+++ b/common/openapi.yml
@@ -223,6 +223,25 @@ paths:
         '500':
           $ref: '#/components/responses/ServerError'
 
+  /api/serviceaccount/v1/{provider}/key:
+    parameters:
+      - $ref: '#/components/parameters/providerParam'
+    get:
+      summary: Returns a service account json key to use to access objects for the provider as the linked account.
+      tags: [ serviceAccount ]
+      operationId: getServiceAccount
+      responses:
+        '200':
+          description: A JSON service account key
+          content:
+            application/json:
+              schema:
+                type: string
+        '404':
+          description: Provider does not support service account keys
+        '500':
+          $ref: '#/components/responses/ServerError'
+
   /passport/v1/validate:
     post:
       summary: Given passports and criteria validate that at least one passport is valid and has a visa that matches at least one crition
--- a/common/openapi.yml
+++ b/common/openapi.yml
@@ -6,7 +6,7 @@ info:
 paths:
   /api/oauth/v1/providers:
     get:
-      summary: Lists the available OIDC providers.
+      summary: Lists the available OAuth providers.
       tags: [ oauth ]
       operationId: listProviders
       responses:
@@ -87,7 +87,6 @@ paths:
           $ref: '#/components/responses/NotFound'
         '500':
           $ref: '#/components/responses/ServerError'
-
   /api/oauth/v1/{provider}/access-token:
     parameters:
       - $ref: '#/components/parameters/providerParam'
@@ -474,9 +473,9 @@ components:
         - ras
         - github
         - fence
-        - dcfFence
+        - dcf-fence
+        - kids-first
         - anvil
-        - kidsFirst
     SshKeyPairType:
       description: Enum containing valid ssh key types.
       type: string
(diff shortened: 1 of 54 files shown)


common/openapi.yml
```diff
@@ -223,6 +223,25 @@ paths:
         '500':
           $ref: '#/components/responses/ServerError'
 
+  /api/fenceAccountKey/v1/{provider}/key:
```
tlangs: if the api group is `fenceAccountKey`, do we really need the `/key` on the end?

service/src/main/java/bio/terra/externalcreds/controllers/FenceAccountKeyController.java
```diff
@@ -0,0 +1,57 @@
...
+
+  @Override
+  public ResponseEntity<String> getFenceAccountKey(Provider provider) {
+    var samUser = samUserFactory.from(request);
+    var maybeLinkedAccount =
+        linkedAccountService.getLinkedAccount(samUser.getSubjectId(), provider);
+    var maybeFenceAccountKey =
+        fenceAccountKeyService.getFenceAccountKey(samUser.getSubjectId(), provider);
+
+    auditLogger.logEvent(
```
tlangs: If the linked account doesn't exist, wouldn't we want to just return a 404? We definitely don't want to log an event if there's no linked account.

service/src/main/java/bio/terra/externalcreds/controllers/FenceAccountKeyController.java
```diff
@@ -0,0 +1,57 @@
...
+    LinkedAccountService linkedAccountService)
+    implements FenceAccountKeyApi {
+
+  @Override
+  public ResponseEntity<String> getFenceAccountKey(Provider provider) {
+    var samUser = samUserFactory.from(request);
+    var maybeLinkedAccount =
+        linkedAccountService.getLinkedAccount(samUser.getSubjectId(), provider);
+    var maybeFenceAccountKey =
+        fenceAccountKeyService.getFenceAccountKey(samUser.getSubjectId(), provider);
```
tlangs: We should probably just `flatMap` the linked account. We could get the userId and provider from the linked account if it exists.

tlangs (changes requested):

common/openapi.yml
samanehsan: That's a good point! I agree it's redundant

service/src/main/java/bio/terra/externalcreds/controllers/FenceAccountKeyController.java
```diff
@@ -0,0 +1,57 @@
...
+    HttpServletRequest request,
+    LinkedAccountService linkedAccountService)
+    implements FenceAccountKeyApi {
+
+  @Override
+  public ResponseEntity<String> getFenceAccountKey(Provider provider) {
+    var samUser = samUserFactory.from(request);
+    var maybeLinkedAccount =
+        linkedAccountService.getLinkedAccount(samUser.getSubjectId(), provider);
+    var response =
```
tlangs: Could we de-nest these `flatMap`s? It gets hard to read when we start nesting. Saving each intermediate `Optional` also improves debuggability. To do that, we can just declare an `AuditLogEvent.Builder` at the top of the method, and then fill in appropriate details as we go, finally `build()`ing and logging the event as we craft the response.

samanehsan:
The tests are failing because the fence providers are not defined in the DB enum. That will get added in https://github.com/DataBiosphere/terra-external-credentials-manager/pull/178!

Shakespeared (approved):

tlangs (approved):

Ghost-in-a-Jar (approved):

PR #183, opened 2024-03-15
ID-1156 Add repo Scope to GitHub Token Requests

tlangs:
Jira: https://broadworkbench.atlassian.net/browse/ID-1156

  \<Don't forget to include the ticket number in the PR title.\>

We need to get GitHub Access Tokens with specific scopes, and those scopes are different from those used to create the OAuth2 link. So, I've split up the scopes into `authorizationScopes` and `tokenScopes`. That way, we can provide the correct scopes at the correct time.


Shakespeared (approved):

salonishah11 (approved):

tlangs:
This ended up not being needed. Turns out, GitHub App tokens _don't have scopes_, and are always scoped to the permissions of the app and the user. Organizations will need to install the app into their orgs in order for the app to get access to private repos.

PR #184, opened 2024-03-28
ID-1126 ECM Retrieves Fence Keys

tlangs:
Jira: https://broadworkbench.atlassian.net/browse/ID-1126

We didn't wire key-retrieval into the get key endpoint! This fixes that, and also adds code to revoke credentials in the fence provider when a link is deleted. Also added tests for that!

### DIFF
--- a/service/src/main/java/bio/terra/externalcreds/controllers/FenceAccountKeyController.java
+++ b/service/src/main/java/bio/terra/externalcreds/controllers/FenceAccountKeyController.java
@@ -6,8 +6,7 @@ import bio.terra.externalcreds.auditLogging.AuditLogger;
 import bio.terra.externalcreds.generated.api.FenceAccountKeyApi;
 import bio.terra.externalcreds.generated.model.Provider;
 import bio.terra.externalcreds.models.FenceAccountKey;
-import bio.terra.externalcreds.services.FenceAccountKeyService;
-import bio.terra.externalcreds.services.LinkedAccountService;
+import bio.terra.externalcreds.services.FenceProviderService;
 import jakarta.servlet.http.HttpServletRequest;
 import java.time.Instant;
 import java.util.Optional;
@@ -18,19 +17,19 @@ import org.springframework.stereotype.Controller;
 public record FenceAccountKeyController(
     AuditLogger auditLogger,
     ExternalCredsSamUserFactory samUserFactory,
-    FenceAccountKeyService fenceAccountKeyService,
-    HttpServletRequest request,
-    LinkedAccountService linkedAccountService)
+    FenceProviderService fenceProviderService,
+    HttpServletRequest request)
     implements FenceAccountKeyApi {
 
   @Override
   public ResponseEntity<String> getFenceAccountKey(Provider provider) {
     var samUser = samUserFactory.from(request);
-    var auditLogEvenBuilder = new AuditLogEvent.Builder()
-        .auditLogEventType(AuditLogEventType.GetServiceAccountKey)
-        .clientIP(request.getRemoteAddr());
+    var auditLogEvenBuilder =
+        new AuditLogEvent.Builder()
+            .auditLogEventType(AuditLogEventType.GetServiceAccountKey)
+            .clientIP(request.getRemoteAddr());
     var maybeLinkedAccount =
-        linkedAccountService.getLinkedAccount(samUser.getSubjectId(), provider);
+        fenceProviderService.getLinkedFenceAccount(samUser.getSubjectId(), provider);
     Optional<FenceAccountKey> maybeFenceAccountKey =
         maybeLinkedAccount.flatMap(
             linkedAccount -> {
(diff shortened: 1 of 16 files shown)


service/src/main/java/bio/terra/externalcreds/services/TokenProviderService.java
```diff
@@ -37,6 +38,7 @@ public TokenProviderService(
         providerTokenClientCache,
         oAuth2Service,
         linkedAccountService,
+        fenceAccountKeyService,
```
samanehsan, line 41: Do we still need the `fenceAccountKeyService` here? It doesn't look like it gets used

tlangs: Its needed for the `super` constructor.

tlangs: Oh, but its never used there either!

service/src/main/java/bio/terra/externalcreds/controllers/FenceAccountKeyController.java
```diff
@@ -18,42 +17,41 @@
...
+    HttpServletRequest request)
     implements FenceAccountKeyApi {
 
   @Override
   public ResponseEntity<String> getFenceAccountKey(Provider provider) {
     var samUser = samUserFactory.from(request);
-    var auditLogEvenBuilder = new AuditLogEvent.Builder()
-        .auditLogEventType(AuditLogEventType.GetServiceAccountKey)
-        .clientIP(request.getRemoteAddr());
+    var auditLogEvenBuilder =
```
samanehsan: Can you fix the typo while you're changing this line? I'm pretty sure I added that 😅 
```suggestion
    var auditLogEventBuilder =
```

tlangs: Done!

service/src/main/java/bio/terra/externalcreds/services/TokenProviderService.java
samanehsan: Oh! Oops those lines were collapsed

service/src/main/java/bio/terra/externalcreds/controllers/OauthApiController.java
```diff
@@ -124,7 +124,7 @@ public ResponseEntity<Void> deleteLink(Provider provider) {
 
     // This can be deleted once there's no longer a dependency on Bond's Datastore
     if (ProviderUtils.isFenceProvider(provider)) {
-      fenceProviderService.deleteFenceLink(samUser.getSubjectId(), provider);
+      fenceProviderService.deleteBondFenceLink(samUser.getSubjectId(), provider);
```
samanehsan, line 127: Why would we only delete the linked account from Bond? I thought it could be in either ECM or Bond once we make the Terra UI talk to ECM 🤔

tlangs: Deleting the linked account in ECM cascade deletes the Fence Account Key in its DB! So, we only need to delete the linked account in ECM. We just _also_ need to tell Bond that we're deleting the linked account

samanehsan: Oh that makes sense!

service/src/main/java/bio/terra/externalcreds/services/FenceProviderService.java
```diff
@@ -117,8 +113,12 @@ public LinkedAccount createLink(
     }
   }
 
-  public void deleteFenceLink(String userId, Provider provider) {
-    bondService.deleteBondLinkedAccount(userId, provider);
+  public void deleteBondFenceLink(String userId, Provider provider) {
+    try {
+      bondService.deleteBondLinkedAccount(userId, provider);
+    } catch (Exception ex) {
```
samanehsan, line 119: What do you think about only logging this warning for non-404 errors?

service/src/test/java/bio/terra/externalcreds/services/FenceProviderServiceTest.java
```diff
@@ -66,11 +87,113 @@ void testGetLinkedFenceAccountKey() {
...
+        .thenReturn(accessTokenResponse);
+
+    when(oAuth2Service.getUserInfo(providerClient, accessTokenResponse.getAccessToken()))
+        .thenReturn(
+            new DefaultOAuth2User(
+                null,
+                Map.of("foo", "bar", provider.getExternalIdClaim(), "barName"),
+                provider.getExternalIdClaim()));
+
+    fenceProviderService.createLink(
```
samanehsan: Can we assert that the created linked account is what we expect?

service/src/test/java/bio/terra/externalcreds/services/ProviderServiceTest.java
```diff
@@ -123,6 +129,139 @@ void testDeleteLinkLinkNotFound() {
...
+                linkedAccount.getUserId(), linkedAccount.getProvider()))
+            .thenReturn(Optional.of(linkedAccount));
+        when(linkedAccountServiceMock.deleteLinkedAccount(
+                linkedAccount.getUserId(), linkedAccount.getProvider()))
+            .thenReturn(true);
+
+        when(providerOAuthClientCacheMock.getProviderClient(linkedAccount.getProvider()))
+            .thenReturn(createClientRegistration(linkedAccount.getProvider()));
+
+        when(oAuth2ServiceMock.authorizeWithRefreshToken(
```
samanehsan, line 235: It looks like these tests mock some of the same things. I think it would be easier to read if those were split out into helper functions, but I don't consider it to be a blocker!

samanehsan (approved):
Thanks for fixing this so quickly! ⚡ I left a few suggestions but they're not blockers

service/src/main/java/bio/terra/externalcreds/services/FenceProviderService.java
tlangs: I don't think its too big of a deal to just log all errors. It would be good to have logs for how many times bond was called to delete a non-existent account. Theoretically, we should see warnings for that increase as time goes on and more accounts exist in ECM only. It would be a good metric to measure ECM's uptake!

service/src/test/java/bio/terra/externalcreds/services/FenceProviderServiceTest.java
tlangs: Done!

service/src/test/java/bio/terra/externalcreds/services/ProviderServiceTest.java
tlangs: Yeah, lots of repeated code. I cleaned it up!

PR #186, opened 2024-04-05
ID-1199 Include proper scopes in token retrieval

tlangs:
Jira: https://broadworkbench.atlassian.net/browse/ID-1199

In prod, we found out that ECM was not providing the proper `google_credentials` scope when getting a token. We don't know why this issue didn't show up in lower envs, but it reared its ugly head once ECM went to prod


rtitle (approved):
Interesting, I never heard of the `google_credentials` scope

PR #189, opened 2024-04-08
VerifyServicePacts

mspector:
Jira: https://broadworkbench.atlassian.net/browse/WM-2554

What:
- Introduces pact verification for ECM's consumers.
- Creates a "provider states enum" for consumers to import ("Provider" is an overloaded term in ECM; I'm talking about Pact providers in this case)

Why:
- Contract testing is something we do!

How:
- Please focus on [this method](https://github.com/DataBiosphere/terra-external-credentials-manager/pull/189/files#diff-9622a5dda2a589d299474341d9bbb46d11e37f2d046f4be8fe693f1a4820a4feR77).
- The new VerifyServicePacts is a SpringBootTest that mocks outer abstraction layers like DAOs and external service clients. It uses SpringBootTest to involve as much of the ECM code as possible while keeping it unit test-like with mocks. This is different from previous verification tests, which have used WebMvcTest to isolate the controller layer, but I think that isolation is actually counter to the spirit of these verifications (I discussed this with @Ghost-in-a-Jar in general terms, but not the specific fact that I'm using a SpringBootTest.)
- Creating the "provider state enum" helps Identiteam maintain control of the states that it supports and can help prevent unlimited growth in the number of states that need to be supported. When writing new contracts, consumers should check which states are already supported, and if necessary, should work together with Identiteam to support new ones for their use case. Identiteam will know best which states make sense to support composability/maintainability.


.java-version
```diff
@@ -1 +1 @@
-17
+17.0
```
mspector, line 1: I needed to do this to get ECM to run on my machine! I will remove it if this isn't generally applicable.

service/build.gradle
```diff
@@ -81,4 +84,23 @@ tasks.withType(Test) {
...
 
-test { maxParallelForks = 1 }
+test {
+	maxParallelForks = 1
+	filter {
+		excludeTestsMatching "VerifyServicePacts*"
+	}
+}
+
+task verifyPacts( type: Test ) {
```
mspector, line 94: this isn't being used anywhere in this PR. I have a followup PR that adds github actions that will run this task when verification is needed, but I want to keep it separate so that Ivan can help review in isolation.

service/src/main/java/bio/terra/externalcreds/pact/ProviderStates.java
```diff
@@ -0,0 +1,10 @@
+package bio.terra.externalcreds.pact;
+
+// Note: This is a regular class, not an Enum, because Java doesn't consider enum values to be
+// constant at compile time, which is a problem when trying to use these states in annotations in
+// TestVerification.java.
+// TODO: confirm that this is a good solution to this problem.
+public class ProviderStates {
+  public static final String USER_IS_REGISTERED = "test_user@test.com is registered with ECM";
+  public static final String ECM_IS_OK = "ECM is ok";
+}
```
mspector, line 10: I mention this in my top-level comment, but the purpose of this "enum" is to make it so that consumers don't have to be copy/pasting plain strings everywhere in order to match ECM provider states. 

E.g. CBAS can import it and use ProviderStates.USER_IS_REGISTERED whenever it wants to call upon that state.

service/src/test/java/bio/terra/externalcreds/pact/VerifyServicePacts.java
```diff
@@ -0,0 +1,125 @@
...
+import org.springframework.test.context.ActiveProfiles;
+import org.springframework.test.context.junit.jupiter.SpringExtension;
+
+@ExtendWith(SpringExtension.class)
+@SpringBootTest(
+    webEnvironment = SpringBootTest.WebEnvironment.DEFINED_PORT,
+    properties = {"server.port=8080"},
+    classes = ExternalCredsWebApplication.class)
+@ActiveProfiles({"test", "human-readable-logging"})
+@au.com.dius.pact.provider.junitsupport.Provider("ecm")
```
mspector, line 51: using the fully qualified name because "ECM providers" is a totally separate concept with the same name!

service/src/test/java/bio/terra/externalcreds/pact/VerifyServicePacts.java
```diff
@@ -0,0 +1,120 @@
...
+  void pactVerificationTestTemplate(PactVerificationContext context) {
+    context.verifyInteraction();
+  }
+
+  @State({ProviderStates.ECM_IS_OK})
+  public void ecmIsOk() {
+    when(statusDAO.isPostgresOk()).thenReturn(true);
+  }
+
+  @State({ProviderStates.USER_IS_REGISTERED})
```
mspector, line 77: Reviewers, please focus on this method. This verification succeeds with my test contracts, but it may not be as strict as you want it.

service/src/main/java/bio/terra/externalcreds/pact/ProviderStates.java
```diff
@@ -0,0 +1,10 @@
+package bio.terra.externalcreds.pact;
+
+// Note: This is a regular class, not an Enum, because Java doesn't consider enum values to be
+// constant at compile time, which is a problem when trying to use these states in annotations in
+// TestVerification.java.
+
+public class ProviderStates {
+  public static final String USER_IS_REGISTERED = "test_user@test.com is registered with ECM";
```
Ghost-in-a-Jar: I think it is best practice for a value like this email to be a provider supplied state value so that we dont have to make sure that both the consumer tests and provider tests have the same hardcoded value. Then this state could be more generic like "a user is registered with ECM".

service/src/test/java/bio/terra/externalcreds/pact/VerifyServicePacts.java
```diff
@@ -0,0 +1,120 @@
...
+        .thenReturn(Optional.of(testAccount));
+
+    when(accessTokenCacheDAO.upsertAccessTokenCacheEntry(isA(AccessTokenCacheEntry.class)))
+        .thenReturn(accessTokenCacheEntry);
+
+    when(mockAccessTokenResponse.getRefreshToken()).thenReturn(testRefreshToken);
+    when(mockAccessTokenResponse.getAccessToken()).thenReturn(testAccessToken);
+
+    when(oAuth2Service.authorizeWithRefreshToken(isA(ClientRegistration.class), any(), any()))
+        .thenReturn(mockAccessTokenResponse);
```
Ghost-in-a-Jar, line 120: I am curious if we can have less mocks setup so we can test more of the actual code, but also this is a good starting point so I am happy to approve!

mspector: yeah, I agree with you - I had trouble getting deeper than this, but this method being mocked doesn't seem to have a lot of interesting logic in it. it occurred to me that someone with deeper knowledge of how it works may have stronger opinions about it

Ghost-in-a-Jar (approved):

service/src/test/java/bio/terra/externalcreds/pact/VerifyServicePacts.java
Shakespeared: I'm also ok approving as a starting point, with hopefully deeper mocking to come later. But this makes sense to me!

Shakespeared (approved):
This LGTM as a starting point! Seems reasonable, we can always update the mocks to be more specific later if we want to

PR #193, opened 2024-04-16
[ID-1144] Configure Sentry

samanehsan:
Jira: https://broadworkbench.atlassian.net/browse/ID-1144

What:
Configure Sentry for ECM

Why:
This will alert us of any unhandled errors that occur 

How:
Use the gradle plugin, and set the sentry config values in the prod spring profile. I was originally going to set the values as env variables in https://github.com/broadinstitute/terra-helmfile/pull/5418 and then use those in application.yaml, but that seemed unnecessary since we have spring profiles.


service/src/main/resources/logback.xml
```diff
@@ -25,6 +25,15 @@
         </encoder>
     </appender>
 
+    <if condition='isDefined("SENTRY_DSN")'>
```
Ghost-in-a-Jar: I think since we are just putting the DSN straight in the providers.ymp (which i think is fine) this condition may never be true? I would just remove it.

Ghost-in-a-Jar (approved):

service/src/main/resources/logback.xml
samanehsan: I thought it would only be true for the production spring profile, but I can double check

Ghost-in-a-Jar: This condition will be looking for an environment variable i think, not something from the spring profile

samanehsan: Oh okay! I'm going to test this out again...I thought I didn't have this set as an env variable when I was running it locally

Shakespeared (approved):

PR #197, opened 2024-04-23
ID-1214 status endpoint improvements

tlangs:
Jira: https://broadworkbench.atlassian.net/browse/ID-1214

Bond checks the status of each fence provider to see if they’re up. This is really useful information, and ECM should do the same.


ichengchang:
@tlangs Your branch is out-of-sync with `dev` branch that's why the verification test failed. This is all fixed now.

common/openapi.yml
```diff
@@ -260,6 +260,19 @@ paths:
...
+      operationId: getStatusDetail
+      security: [ ]
+      responses:
+        '200':
+          $ref: '#/components/responses/SystemStatusDetailResponse'
+        '500':
+          $ref: '#/components/responses/ServerError'
+        '503':
+          $ref: '#/components/responses/SystemStatusDetailResponse'
   /status:
```
samanehsan, line 276: Should we mark the old `status` endpoint as deprecated?

tlangs: Not a bad idea!

tlangs: Done!

service/src/main/java/bio/terra/externalcreds/services/StatusServiceCache.java
```diff
@@ -0,0 +1,81 @@
...
+import org.springframework.cache.annotation.Cacheable;
+import org.springframework.scheduling.annotation.Scheduled;
+import org.springframework.stereotype.Component;
+
+/**
+ * Creating a provider client requires an api call to [provider
+ * issuer]/.well-known/openid-configuration. That information almost never changes and we don't want
+ * to hammer that api. Therefore this cache.
+ *
+ * <p>The cache is reset every 6 hours to detect infrequent changes.
```
samanehsan: I'm wondering if 6 hours is too long for caching Sam's status, since the ECM status relies on Sam being up 🤔

samanehsan: Oh I realized this was copied over from `ProviderOAuthClientCache` and it's actually set to 1 minute here for Sam and 5 minutes for the providers. Could you update this comment?

service/src/test/java/bio/terra/externalcreds/services/StatusServiceCacheTest.java
```diff
@@ -0,0 +1,134 @@
...
+    var status1 = statusServiceCache.getProviderStatus(Provider.GITHUB);
+    var status2 = statusServiceCache.getProviderStatus(Provider.GITHUB);
+
+    // Assert
+    assertEquals(status1, status2);
+    verify(providerOAuthClientCache, times(1)).getProviderClient(Provider.GITHUB);
+  }
+
+  @Test
+  void getGetsSuccessfulSamStatus() throws ApiException {
```
samanehsan: Tiny renaming suggestion:
```suggestion
  void testGetsSuccessfulSamStatus() throws ApiException {
```

service/src/test/java/bio/terra/externalcreds/services/StatusServiceTest.java
```diff
@@ -0,0 +1,94 @@
...
+import org.springframework.boot.test.mock.mockito.MockBean;
+
+class StatusServiceTest extends BaseTest {
+
+  @Autowired private StatusService statusService;
+  @MockBean private StatusServiceCache statusServiceCache;
+  @MockBean private StatusDAO statusDAO;
+
+  @Test
+  void testgetSystemStatusDetail() {
```
samanehsan, line 22: One more nit-pick 😅 
```suggestion
  void testGetSystemStatusDetail() {
```

service/src/main/java/bio/terra/externalcreds/services/StatusServiceCache.java
samanehsan: Also if the provider cache is reset every 6 hours, maybe checking it every 5 minutes is too often?

samanehsan:
After this is merged we should update the externalcreds chart to use this new status endpoint! Is there a separate ticket for that?

service/src/main/java/bio/terra/externalcreds/services/StatusServiceCache.java
```diff
@@ -32,7 +28,6 @@ public StatusServiceCache(
   // Cache the status of each provider.
   // This needs to happen so that if a providerClient fails to get created, we don't hammer the
   // provider with requests to our own status endpoint.
```
samanehsan, line 30: We can remove this comment about caching here too

samanehsan (approved):
Lgtm! ☀️

PR #203, opened 2024-07-12
ID-1130 Store NIH Username in ECM

tlangs:

We're starting to move away from Thurloe! One of the things Thurloe stores is the user's NIH username. This will be migrated to ECM in the future when Shibboleth is retired in favor of OAuth2, but for now, we can make Orchestration talk to ECM instead of Thurloe. 

This PR introduces the NIH Account table, which just stores NIH Usernames and link expiration times. API endpoints have been added to support Orch's functionality, as outlined [here](https://docs.google.com/document/d/1cdrALm4o1NZvFg4iE2eqDepKpwL2HPfY2N33A3j2--k/edit#heading=h.naxdnn42df8w).

### DIFF
--- a/common/openapi.yml
+++ b/common/openapi.yml
@@ -4,6 +4,85 @@ info:
   description: A manager for external credentials
   version: 0.0.1
 paths:
+  /api/nih/v1/account:
+    get:
+      summary: Get the NIH account for the user
+      tags: [ nihAccount ]
+      operationId: getNihAccount
+      responses:
+        '200':
+          description: The user's NIH account
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/NihAccountModel'
+        '404':
+          description: User does not have a linked NIH account
+        '500':
+          $ref: '#/components/responses/ServerError'
+    put:
+      summary: Link the user's NIH account
+      tags: [ nihAccount ]
+      operationId: linkNihAccount
+      requestBody:
+        content:
+          'application/json':
+            schema:
+              $ref: '#/components/schemas/NihAccountModel'
+      responses:
+          '202':
+            description: NIH account linked
+          '400':
+            $ref: '#/components/responses/BadRequest'
+          '403':
+            $ref: '#/components/responses/PermissionDenied'
+          '500':
+            $ref: '#/components/responses/ServerError'
+  /api/admin/nih/v1/userForUsername/{nih_username}:
+    parameters:
+      - name: nih_username
+        in: path
+        required: true
+        schema:
+          type: string
+    get:
+      summary: Get the NIH Account for the NIH username
+      tags: [ nihAccount ]
+      operationId: getNihAccountForUsername
+      responses:
+        '200':
+          description: The NIH Account for the NIH username
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/NihAccountModel'
+        '403':
+          $ref: '#/components/responses/PermissionDenied'
+        '404':
+          description: User not found for the NIH username
+        '500':
+          $ref: '#/components/responses/ServerError'
+  /api/admin/nih/v1/activeAccounts:
+    get:
+      summary: Get all active NIH Accounts
+      tags: [ nihAccount ]
+      operationId: getActiveNihAccounts
+      responses:
+        '200':
+          description: The user for the NIH username
+          content:
+            application/json:
+              schema:
+                type: array
+                items:
+                  $ref: '#/components/schemas/NihAccountModel'
+        '403':
+          $ref: '#/components/responses/PermissionDenied'
+        '404':
+          description: User not found for the NIH username
+        '500':
+          $ref: '#/components/responses/ServerError'
+
   /api/oauth/v1/providers:
     get:
       summary: Lists the available OAuth providers.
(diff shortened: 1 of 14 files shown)


service/src/main/java/bio/terra/externalcreds/controllers/NihAccountApiController.java
```diff
@@ -0,0 +1,58 @@
...
+import bio.terra.externalcreds.generated.model.NihAccountModel;
+import bio.terra.externalcreds.services.NihAccountService;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import jakarta.servlet.http.HttpServletRequest;
+import java.util.List;
+import org.springframework.http.ResponseEntity;
+import org.springframework.stereotype.Controller;
+
+@Controller
+public record NihAccountApiController(
```
Ghost-in-a-Jar, line 15: just out of curiosity why is this a record and not just a normal class?

service/src/main/java/bio/terra/externalcreds/dataAccess/NihAccountDAO.java
```diff
@@ -0,0 +1,108 @@
...
+  final NamedParameterJdbcTemplate jdbcTemplate;
+
+  public NihAccountDAO(NamedParameterJdbcTemplate jdbcTemplate) {
+    this.jdbcTemplate = jdbcTemplate;
+  }
+
+  @WithSpan
+  public Optional<NihAccount> getNihAccount(String userId) {
+    var namedParameters = new MapSqlParameterSource().addValue("userId", userId);
+    var query = "SELECT * FROM nih_account WHERE user_id = :userId";
```
Ghost-in-a-Jar: imo better to list out the columns explicitly in a DAO

service/src/main/java/bio/terra/externalcreds/dataAccess/NihAccountDAO.java
```diff
@@ -0,0 +1,108 @@
...
+        DataAccessUtils.singleResult(
+            jdbcTemplate.query(query, namedParameters, NIH_ACCOUNT_ROW_MAPPER)));
+  }
+
+  @WithSpan
+  public List<NihAccount> getExpiredNihAccounts() {
+    var namedParameters =
+        new MapSqlParameterSource("expirationCutoff", Timestamp.from(Instant.now()));
+    var query =
+        "SELECT DISTINCT na.* FROM nih_account na" + " WHERE na.expires < :expirationCutoff";
```
Ghost-in-a-Jar: distinct is sort of a smell imo, is there a way to do this with a group by? also listing out the columns explicitly would be nice

service/src/main/resources/db/changelog/changesets/20240711_add_nih_account_table.yaml
```diff
@@ -0,0 +1,31 @@
...
+                    primaryKey: true
+                    nullable: false
+              - column:
+                  name: user_id
+                  type: text
+                  constraints:
+                    nullable: false
+                    unique: true
+              - column:
+                  name: nih_username
```
Ghost-in-a-Jar, line 23: maybe index this?

tlangs: oh good idea!

service/src/main/java/bio/terra/externalcreds/controllers/NihAccountApiController.java
tlangs: Because it was copy-pasted from another controller, which was also a record. 🤷‍♂️

common/openapi.yml
```diff
@@ -4,6 +4,85 @@ info:
...
+        '500':
+          $ref: '#/components/responses/ServerError'
+  /api/admin/nih/v1/activeAccounts:
+    get:
+      summary: Get all active NIH Accounts
+      tags: [ nihAccount ]
+      operationId: getActiveNihAccounts
+      responses:
+        '200':
+          description: The user for the NIH username
```
Shakespeared, line 72: Is this description accurate? It sounds like this endpoint actually returns a list of all active NIH accounts

common/openapi.yml
```diff
@@ -4,6 +4,85 @@ info:
...
+          content:
+            application/json:
+              schema:
+                type: array
+                items:
+                  $ref: '#/components/schemas/NihAccountModel'
+        '403':
+          $ref: '#/components/responses/PermissionDenied'
+        '404':
+          description: User not found for the NIH username
```
Shakespeared, line 82: same question here, seems like this might be left over from the previous endpoint and it should be more like "no active NIH accounts found"

Shakespeared (approved):
approved with a couple comments where I think you just need to update the API descriptions

service/src/main/java/bio/terra/externalcreds/dataAccess/NihAccountDAO.java
tlangs: Yeah, good point. Copy-pasta-ing from other methods sure is catching me out 😅

service/src/main/java/bio/terra/externalcreds/dataAccess/NihAccountDAO.java
tlangs: I don't actually think we need to `distinct` or `group by` here at all. The unique constrains on the table should take care of that.

PR #208, opened 2024-07-17
ID-1330 Admin Endpoints for Orchestration

tlangs:

What:

Some Admin endpoints for Orchestration so that it can get Terra Users from NIH Usernames as a part of the NBCIAccess Allowlist Sync process.

Why:

This enables Orchestration to talk to ECM instead of Thurloe to get Terra Users who have a linked NIH Account

How:

The accounts that are allowed to call the admin endpoints are specified in the config. Every request to the Admin endpoints is checked against that list of email addresses to make sure the caller is authorized.


service/src/main/java/bio/terra/externalcreds/controllers/AdminApiController.java
```diff
@@ -0,0 +1,46 @@
...
+  @Override
+  public ResponseEntity<AdminLinkInfo> getLinkedAccountForExternalId(
+      Provider provider, String externalId) {
+    requireAdmin();
+    var linkedAccount = linkedAccountService.getLinkedAccountForExternalId(provider, externalId);
+    return ResponseEntity.of(linkedAccount.map(OpenApiConverters.Output::convertAdmin));
+  }
+
+  private void requireAdmin() {
+    var samUser = samUserFactory.from(request);
```
Ghost-in-a-Jar, line 41: does this actually reach out to sam or just parse the request for sam user info?

service/src/main/resources/db/changelog/changesets/20240711_add_nih_account_table.yaml
```diff
@@ -0,0 +1,38 @@
...
+                  type: timestamp
+                  constraints:
+                    nullable: false
+        - createIndex:
+            columns:
+              - column:
+                  descending: true
+                  name: nih_username
+            indexName: idx_nih_username
+            tableName: nih_account
```
Ghost-in-a-Jar: It looks like this table isnt being used in this pr (linked_account instead), why are we making this new table then?

service/src/main/java/bio/terra/externalcreds/controllers/AdminApiController.java
tlangs: It gets the Sam user using the token in the request.

service/src/main/resources/db/changelog/changesets/20240711_add_nih_account_table.yaml
tlangs: Whoops, bad rebase.

Ghost-in-a-Jar (approved):

PR #209, opened 2024-07-25
ID-1330 Orchestration can PUT eRA Commons accounts

tlangs:
Jira:  https://broadworkbench.atlassian.net/browse/ID-1330

Since RAS is on the slow-end of getting us the correct redirects for OAuth login with eRA Commons credentials, we should just have ECM store the result of the SAML handshake. This PR lets Orch `PUT` a linked account for eRA Commons. It also makes getting a token from a linked eRA Account throw an `UnsupportedOperationException`, since we don't have a real refresh token to get an access token with.

### DIFF
--- a/common/openapi.yml
+++ b/common/openapi.yml
@@ -4,7 +4,29 @@ info:
   description: A manager for external credentials
   version: 0.0.1
 paths:
-  /api/admin/{provider}/v1/userForExternalId/{externalId}:
+  /api/admin/v1/{provider}:
+    parameters:
+      - $ref: '#/components/parameters/providerParam'
+    put:
+      summary: Link a user to an external account with a fake refresh token
+      tags: [ admin ]
+      operationId: putLinkedAccountWithFakeToken
+      requestBody:
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/AdminLinkInfo'
+        required: true
+      responses:
+        '204':
+          description: Linked Account Created
+        '400':
+          $ref: '#/components/responses/BadRequest'
+        '403':
+          $ref: '#/components/responses/PermissionDenied'
+        '500':
+          $ref: '#/components/responses/ServerError'
+  /api/admin/v1/{provider}/userForExternalId/{externalId}:
     parameters:
       - $ref: '#/components/parameters/providerParam'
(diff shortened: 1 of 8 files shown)


common/openapi.yml
```diff
@@ -4,7 +4,29 @@ info:
...
+      responses:
+        '204':
+          description: Linked Account Created
+        '400':
+          $ref: '#/components/responses/BadRequest'
+        '403':
+          $ref: '#/components/responses/PermissionDenied'
+        '500':
+          $ref: '#/components/responses/ServerError'
+  /api/admin/v1/{provider}/userForExternalId/{externalId}:
```
tlangs, line 29: moved the api version to the correct place.

service/src/main/java/bio/terra/externalcreds/controllers/AdminApiController.java
```diff
@@ -21,6 +23,23 @@ public record AdminApiController(
...
+  @Override
+  public ResponseEntity<Void> putLinkedAccountWithFakeToken(
+      Provider provider, AdminLinkInfo adminLinkInfo) {
+    requireAdmin();
+    var linkedAccount =
+        new LinkedAccount.Builder()
+            .isAuthenticated(true)
+            .provider(provider)
+            .userId(adminLinkInfo.getUserId())
+            .refreshToken("fake-refresh-token")
```
tlangs, line 35: I've made it so that trying to get an access token from an eRA Commons account fails with a helpful message instead of some esoteric permissions error.

service/src/test/java/bio/terra/externalcreds/controllers/AdminApiControllerTest.java
```diff
@@ -66,9 +111,9 @@ void testGetLinkedAccountForExternalIdNonAdmin() throws Exception {
       var inputLinkedAccount = TestUtils.createRandomLinkedAccount(Provider.ERA_COMMONS);
 
       mvc.perform(
-              get("/api/admin/"
+              get("/api/admin/v1/"
```
tlangs, line 114: Result of putting the API Version in the right place.

Ghost-in-a-Jar (approved):

PR #210, opened 2024-07-25
ID-1330 Orch can DELETE linked eRA COmmons accounts

tlangs:

Forgot to add the DELETE ability. Also, added a guard so that _only_ eRA Commons accounts can be linked with a fake token. We shouldn't even allow an admin to make a link for any other provider, since they all support OAuth.

### DIFF
--- a/common/openapi.yml
+++ b/common/openapi.yml
@@ -8,7 +8,7 @@ paths:
     parameters:
       - $ref: '#/components/parameters/providerParam'
     put:
-      summary: Link a user to an external account with a fake refresh token
+      summary: Link a user to an external account with a fake refresh token. For eRA Commons only.
       tags: [ admin ]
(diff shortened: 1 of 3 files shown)


common/openapi.yml
```diff
@@ -26,6 +26,28 @@ paths:
...
+    delete:
+      summary: Unlink a user from an external account
+      tags: [ admin ]
+      operationId: adminDeleteLinkedAccount
+      requestBody:
+        content:
+          text/plain:
+            schema:
+              type: string
+              description: The Sam User Id for a user
```
tlangs, line 38: I'm unsure whether to have this be plain-text or a json object with a `userId` field. If you've got thoughts, lemme know!

Ghost-in-a-Jar (approved):

PR #216, opened 2024-10-22
[CORE-108] Add DUOS urls to era-commons allowed redirectUris

samanehsan:
**Jira:** https://broadworkbench.atlassian.net/browse/CORE-108

**What:** Add DUOS to list of allowed era-commons redirect URIs. Note that this is for non-prod environments where era-commons is enabled.

**Why:** Arbitrary redirect uris are not allowed due to security considerations. When the oauth provider does the redirect it includes information to allow the user to link their account and retrieve an access token.

**How:** Update providers.yaml

### DIFF
--- a/service/src/main/resources/providers.yml
+++ b/service/src/main/resources/providers.yml
@@ -97,7 +97,9 @@ externalcreds:
                                     "https://dev.terra.biodatacatalyst.nhlbi.nih.gov/ecm-callback",
                                     "https://terra.dsde-dev.broadinstitute.org/ecm-callback" ]
     era-commons:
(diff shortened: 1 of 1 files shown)


marctalbott (approved):
I have limited ECM knowledge, but this seems reasonable given the PR and ticket description.

rjohanek (approved):
seems reasonable to me

fboulnois (approved):
URLs look good 👍

rushtong:
~I don't see the prod urls in this PR, should that happen here or separately?~
Never mind - I see in the ticket description that this is not enabled for prod yet.

rushtong (approved):
👍🏽

PR #217, opened 2024-10-24
CORE-108: Add local DUOS url with port

rushtong:
**Jira**: https://broadworkbench.atlassian.net/browse/CORE-108

**What**: Add local version of DUOS to list of allowed era-commons redirect URIs.

**Why**: Arbitrary redirect uris are not allowed due to security considerations. When the oauth provider does the redirect it includes information to allow the user to link their account and retrieve an access token. Note that this is related to https://github.com/DataBiosphere/terra-external-credentials-manager/pull/216

**How**: Update providers.yaml

### DIFF
--- a/service/src/main/resources/providers.yml
+++ b/service/src/main/resources/providers.yml
@@ -99,6 +99,7 @@ externalcreds:
     era-commons:
       allowedRedirectUriPatterns: [ "https://bvdp-saturn-dev.appspot.com/ecm-callback",
                                     "https://localhost.dsde-dev.broadinstitute.org/",
+                                    "https://localhost.dsde-dev.broadinstitute.org:3000/",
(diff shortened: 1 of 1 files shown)


service/src/main/resources/providers.yml
```diff
@@ -99,6 +99,7 @@ externalcreds:
     era-commons:
       allowedRedirectUriPatterns: [ "https://bvdp-saturn-dev.appspot.com/ecm-callback",
                                     "https://localhost.dsde-dev.broadinstitute.org/",
+                                    "https://localhost.dsde-dev.broadinstitute.org:3000/",
```
rushtong, line 102: Note to reviewers, we use both of these urls. When running as a docker image, the app runs on port 80. When running via `npm start`, the app runs on port 3000

samanehsan (approved):

kevinmarete (approved):

PR #218, opened 2024-10-24
[CORE-110] Store additional info in oauth state param

samanehsan:
Jira: https://broadworkbench.atlassian.net/browse/CORE-110

What:
Add a new POST authorization-url endpoint that allows users of ECM to include additional information in the oauth state parameter (which will persist after the user goes through the oauth flow). The information in the state will be returned in the createLink endpoint response.

Why:
This will allow a DUOS to specify a "redirectTo" location when a user links their NIH account.

How:
- Add a POST authorization-url endpoint that accepts arbitrary keys/values in the request body and includes them in the oauth state
Screenshot 2024-10-31 at 2 22 55 PM

- After ECM creates a linked account, decode the state and add `additionalState` to the response
Screenshot 2024-10-31 at 2 24 17 PM

Testing:
In addition to unit testing, I ran ECM locally and tested the endpoints using the Github provider.

### DIFF
--- a/common/openapi.yml
+++ b/common/openapi.yml
@@ -136,6 +136,7 @@ paths:
     parameters:
       - $ref: '#/components/parameters/providerParam'
       - $ref: '#/components/parameters/redirectUriParam'
+      - $ref: '#/components/parameters/additionalStateParam'
     get:
       summary: Builds an OAuth authorization URL that a user must use to initiate the OAuth dance.
       tags: [ oauth ]
@@ -246,6 +247,7 @@ paths:
     parameters:
       - $ref: '#/components/parameters/providerParam'
       - $ref: '#/components/parameters/redirectUriParam'
+      - $ref: '#/components/parameters/additionalStateParam'
     get:
       summary: Builds an OAuth authorization URL that a user must use to initiate the OAuth dance.
       tags: [ oidc ]
@@ -425,6 +427,15 @@ components:
       example: "http://localhost:9000/fence-callback"
       schema:
         type: string
+    additionalStateParam:
+      name: additionalState
+      description: Information to include in the oidc authorization state in the form of key/value pairs
+      in: query
+      required: false
+      content:
+        application/json:
+          schema:
+            type: object
     stateParam:
       name: state
       description: oidc authorization state, must be identical to state embedded in response from getAuthUrl
--- a/common/openapi.yml
+++ b/common/openapi.yml
@@ -541,6 +541,8 @@ components:
           format: date-time
         authenticated:
           type: boolean
+        additionalState:
+          type: object
     ErrorReport:
       type: object
       required: [ message, statusCode ]
--- a/common/openapi.yml
+++ b/common/openapi.yml
@@ -526,6 +526,8 @@ components:
       type: object
       additionalProperties:
         type: string
+      example: |
+        {"redirectTo": "https://my.redirect.url/"}
     PassportProvider:
       description: Enum containing valid passport providers.
       type: string
--- a/common/openapi.yml
+++ b/common/openapi.yml
@@ -136,12 +136,30 @@ paths:
     parameters:
       - $ref: '#/components/parameters/providerParam'
       - $ref: '#/components/parameters/redirectUriParam'
-      - $ref: '#/components/parameters/additionalStateParam'
     get:
       summary: Builds an OAuth authorization URL that a user must use to initiate the OAuth dance.
       tags: [ oauth ]
       operationId: getAuthorizationUrl
       description: First step to link a new provider. Creates a 1 time use link. There can only be 1 active link per user per provider. The last link created will invalidate any prior links.
+      responses:
+        '200':
+          description: The authorization URL
+          content:
+            text/plain:
+              schema:
+                type: string
+        '500':
+          $ref: '#/components/responses/ServerError'
+    post:
+      summary: Builds an OAuth authorization URL that a user must use to initiate the OAuth dance and includes any additional params in the state.
+      tags: [ oauth ]
+      operationId: getAuthorizationUrlWithAdditionalParams
+      description: First step to link a new provider. Creates a 1 time use link. There can only be 1 active link per user per provider. The last link created will invalidate any prior links.
+      requestBody:
+        content:
+          'application/json':
+            schema:
+              $ref: '#/components/schemas/GetAuthUrlRequest'
       responses:
         '200':
           description: The authorization URL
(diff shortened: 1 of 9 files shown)


common/openapi.yml
```diff
@@ -136,6 +136,7 @@ paths:
     parameters:
       - $ref: '#/components/parameters/providerParam'
       - $ref: '#/components/parameters/redirectUriParam'
+      - $ref: '#/components/parameters/additionalStateParam'
```
davidangb: This has the same issue discussed at https://broadinstitute.slack.com/archives/C07QQ3L0AHX/p1730216050615079 … any consumers of the ECM client will see this as a breaking change when they update to the latest client version, and will need to update their calling code.

I think we should move forward with it, but we'll need to notify any users of the client. Possibly that's only our team???

common/openapi.yml
```diff
@@ -425,6 +427,15 @@ components:
...
         type: string
+    additionalStateParam:
+      name: additionalState
+      description: Information to include in the oidc authorization state in the form of key/value pairs
+      in: query
+      required: false
+      content:
+        application/json:
+          schema:
+            type: object
```
davidangb: consider using `additionalProperties` here. As-is, the ECM generated client will have a method of:
```java
public String getAuthorizationUrl(Provider provider, String redirectUri, Object additionalState)
```

if you use
```yaml
          schema:
            type: object
            additionalProperties: true
```
it will result in a nicer signature of:
```java
public String getAuthorizationUrl(Provider provider, String redirectUri, Map<String, Object> additionalState)
```

and if you wanted to be really opinionated and say that the values in the KVPs can only be strings, you could use
```yaml
          schema:
            type: object
            additionalProperties:
              type: string
```
resulting in a signature of:
```java
public String getAuthorizationUrl(Provider provider, String redirectUri, Map<String, String> additionalState)
```

I think it'll be a lot nicer for callers to work with a `Map` than a raw `Object`.

This will cascade a bunch of changes through the server-side controllers and model objects, too :(

render_config.sh
```diff
@@ -54,7 +54,7 @@ fi
   echo export SAM_ADDRESS=https://sam.dsde-${ENV}.broadinstitute.org
 } >> "${SECRET_ENV_VARS_LOCATION}"
 
-$VAULT_COMMAND -field=swagger-client-id "$ECM_VAULT_PATH/swagger-client-id" >"$SERVICE_OUTPUT_LOCATION/swagger-client-id"
+gcloud secrets versions access latest --secret=externalcreds-swagger-client-id --project=broad-dsde-dev | jq -r '."swagger-client-id"' >"$SERVICE_OUTPUT_LOCATION/swagger-client-id"
```
davidangb: the rest of render_config.sh looks like it supports multiple environments; I see `if [ $ENV == 'prod' ];` clauses. Does this need to support more than the hardcoded `broad-dsde-dev`?

marctalbott (approved):

davidangb (commented):
See comment inline about using a `Map` instead of a raw `Object` for the additional state. I think this would be nicer for callers - and probably nicer for the internal code - but would drive a bunch of changes.

common/openapi.yml
samanehsan: Good point! The Terra UI is the only service that calls this endpoint (that I know of)

common/openapi.yml
samanehsan: I agree that looks easier to work with than a generic Object!

samanehsan: I'm getting this error when I try to change the object param into a Map<String, String> 😞 "Failed to convert value of type 'java.lang.String' to required type 'java.util.Map'; Cannot convert value of type 'java.lang.String' to required type 'java.util.Map' for property 'additionalState': no matching editors or conversion strategy found". Do you have any ideas what I should try?

Here is the schema:
```
    additionalStateParam:
      name: additionalState
      description: Information to include in the oidc authorization state in the form of key/value pairs
      in: query
      required: false
      content:
        application/json:
          schema:
            type: object
            additionalProperties:
              type: string
```

And here is the url param: `additionalState=%7B%0A%20%20%22additionalProp1%22%3A%20%22string%22%2C%0A%20%20%22additionalProp2%22%3A%20%22string%22%2C%0A%20%20%22additionalProp3%22%3A%20%22string%22%0A%7D`

render_config.sh
samanehsan: I've only ever used this for local development but the script implies otherwise, so I made the google project reflect the environment

common/openapi.yml
```diff
@@ -503,6 +522,10 @@ components:
             linkExpireTime:
               type: string
               format: date-time
+    GetAuthUrlRequest:
+      type: object
+      additionalProperties:
+        type: string
```
davidangb, line 528: nit, and optional: provide an example here to make the swagger-ui a little prettier, like
```
    GetAuthUrlRequest:
      type: object
      additionalProperties:
        type: string
      example: |
        {"redirectTo": "https://my.redirect.url/"}
```

render_config.sh
```diff
@@ -25,36 +25,38 @@ if [ -f "${SECRET_ENV_VARS_LOCATION}" ]; then
...
+  echo export ANVIL_CLIENT_SECRET="$(gcloud secrets versions access latest --secret=externalcreds-fence --project="${GOOGLE_PROJECT}" | jq -r '."anvil-client-secret"')"
+  echo export FENCE_CLIENT_ID="$(gcloud secrets versions access latest --secret=externalcreds-fence --project="${GOOGLE_PROJECT}" | jq -r '."client-id"')"
+  echo export FENCE_CLIENT_SECRET="$(gcloud secrets versions access latest --secret=externalcreds-fence --project="${GOOGLE_PROJECT}" | jq -r '."client-secret"')"
+  echo export DCF_FENCE_CLIENT_ID="$(gcloud secrets versions access latest --secret=externalcreds-fence --project="${GOOGLE_PROJECT}" | jq -r '."dcf-fence-client-id"')"
+  echo export DCF_FENCE_CLIENT_SECRET="$(gcloud secrets versions access latest --secret=externalcreds-fence --project="${GOOGLE_PROJECT}" | jq -r '."dcf-fence-client-secret"')"
+  echo export KIDS_FIRST_CLIENT_ID="$(gcloud secrets versions access latest --secret=externalcreds-fence --project="${GOOGLE_PROJECT}" | jq -r '."kids-first-client-id"')"
+  echo export KIDS_FIRST_CLIENT_SECRET="$(gcloud secrets versions access latest --secret=externalcreds-fence --project="${GOOGLE_PROJECT}" | jq -r '."kids-first-client-secret"')"
```
davidangb, line 53: wow, I didn't realize how much in this file was hardcoded to broad-dsde-dev!

davidangb (approved):
looks good, thanks for bearing with me! One optional suggestion inline; giving this a proactive 👍

PR #219, opened 2024-10-25
CORE-108: Fix local duos instance domains

rushtong:
**Jira**: https://broadworkbench.atlassian.net/browse/CORE-108

**What**: Fix local versions of DUOS' list of allowed era-commons redirect URIs.

**Why**: Arbitrary redirect uris are not allowed due to security considerations. When the oauth provider does the redirect it includes information to allow the user to link their account and retrieve an access token. Note that this is related to https://github.com/DataBiosphere/terra-external-credentials-manager/pull/216 and https://github.com/DataBiosphere/terra-external-credentials-manager/pull/217

**How**: Update providers.yaml


fboulnois (approved):
good catch 👍

kevinmarete (approved):

rjohanek (approved):

PR #241, opened 2025-02-11
DT-659: Update RAS Redirects for DUOS

rushtong:
Jira: https://broadworkbench.atlassian.net/browse/DT-659

This PR adds the current set of DUOS eRA Commons redirect urls to RAS so DUOS can switch over to using RAS authentication instead of eRA Commons.

### DIFF
--- a/service/src/main/resources/providers.yml
+++ b/service/src/main/resources/providers.yml
@@ -95,7 +95,10 @@ externalcreds:
                                     "http://localhost:3000/ecm-callback",
                                     "https://dev.ukbiobank.terra.bio/ecm-callback",
                                     "https://dev.terra.biodatacatalyst.nhlbi.nih.gov/ecm-callback",
-                                    "https://terra.dsde-dev.broadinstitute.org/ecm-callback" ]
+                                    "https://terra.dsde-dev.broadinstitute.org/ecm-callback",
+                                    "https://local.dsde-dev.broadinstitute.org",
+                                    "https://local.dsde-dev.broadinstitute.org:3000",
+                                    "https://duos-k8s.dsde-dev.broadinstitute.org"]
     era-commons:
       allowedRedirectUriPatterns: [ "https://bvdp-saturn-dev.appspot.com/ecm-callback",
                                     "https://local.dsde-dev.broadinstitute.org",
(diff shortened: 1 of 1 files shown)


service/src/main/resources/providers.yml
```diff
@@ -136,10 +139,11 @@ spring.config.activate.on-profile: 'staging'
 externalcreds:
   providers:
     ras:
-      allowedRedirectUriPatterns: [ "https://bvdp-saturn-staging.appspot.com/ecm-callback" ]
+      allowedRedirectUriPatterns: [ "https://bvdp-saturn-staging.appspot.com/ecm-callback",
+                                    "https://duos-k8s.dsde-staging.broadinstitute.org"]
     era-commons:
       allowedRedirectUriPatterns: [ "https://bvdp-saturn-staging.appspot.com/ecm-callback",
-                                    "https://duos-k8s.dsde-staging.broadinstitute.org/" ]
+                                    "https://duos-k8s.dsde-staging.broadinstitute.org" ]
```
rushtong, line 146: Drive by fix to keep all of the url formats consistent across environments.

service/src/main/resources/providers.yml
```diff
@@ -95,7 +95,10 @@ externalcreds:
                                     "http://localhost:3000/ecm-callback",
                                     "https://dev.ukbiobank.terra.bio/ecm-callback",
                                     "https://dev.terra.biodatacatalyst.nhlbi.nih.gov/ecm-callback",
-                                    "https://terra.dsde-dev.broadinstitute.org/ecm-callback" ]
+                                    "https://terra.dsde-dev.broadinstitute.org/ecm-callback",
+                                    "https://local.dsde-dev.broadinstitute.org",
```
pshapiro4broad, line 99: There's a case where we running local consent using port 80?

pshapiro4broad (approved):
Looks OK to me, as far as I can tell

service/src/main/resources/providers.yml
rushtong: Yes, if we run the UI in a [docker compose](https://github.com/DataBiosphere/duos-ui/blob/develop/docker-compose.yaml), that's on port 80.

samanehsan (approved):

PR #253, opened 2025-03-05
CORE-332: disable fence

davidangb:

## What:
* Fence service account support is removed from ECM to stop calling Gen3
* ECM returns a 404 to any downstream services requesting a fence service account key
* Deleted (lots of) implementation code which previously supported fence keys

## Why:
Gen3 wants to stop supporting this feature on their end; therefore, ECM needs to stop calling Gen3.

## Notes:
I may have missed some implementation code which can still be deleted. Do you see anything else that should be deleted? This PR is already removing quite a lot.


service/src/main/java/bio/terra/externalcreds/services/FenceProviderService.java
```diff
@@ -19,32 +18,22 @@
 @Slf4j
 public class FenceProviderService extends ProviderService {
```
dvoet: I think this class can go away and use `TokenProviderService` instead.

Bonus points: merge `TokenProviderService` with `PassportProviderService`, remove the big switch statement in `OauthApiController.createLink`, drive the decision about whether or not to get a passport from either a new config setting for a provider or from the existence of `ga4gh_passport_v1` scope.

kevinmarete (approved):

dvoet (changes requested):
1 inline comment

also in service/src/main/resources/providers.yml, remove instances of `google_credentials` scope

davidangb:
@dvoet I have removed `google_credentials` scopes in provider config, and removed `FenceProviderService` entirely in favor of `TokenProviderService`.

I am game to tackle merging `TokenProviderService` with `PassportProviderService` … but how about doing that in a separate PR, since this one is already pretty big?

dvoet (approved):

PR #255, opened 2025-03-05
[CORE-223] Rework VisaCriterion inheritance to fix ValidatePassportResult deserialization issue

marctalbott:
Jira: [CORE-223](https://broadworkbench.atlassian.net/browse/CORE-223)

**What:**
* Make `RASv1Dot1VisaCriterion` a subtype of `VisaCriterion` and reference `VisaCriterion` directly instead of `RASv1Dot1VisaCriterion`.
* Breaking change: `OneOfValidatePassportRequestCriteriaItems` and `OneOfValidatePassportResultMatchedCriterion` are no longer part of the generated client. Replace usage with either `VisaCriterion` or `RASv1Dot1VisaCriterion` as needed.

**Why:**
TDR was failing to deserialize the `ValidatePassportResult` response from ECM because of a bug in the generated ECM model classes caused by the change introduced in https://github.com/swagger-api/swagger-codegen-generators/pull/1259 which was pulled into ECM when upgrading `swagger-codegen-cli` past 3.0.54 (https://github.com/DataBiosphere/terra-external-credentials-manager/pull/227, ECM `v1.63.0`). Before this change, the class representing the validation criterion looked like this:

```
@JsonTypeInfo(
  use = Id.NAME,
  include = As.PROPERTY,
  property = "type"
)
@JsonSubTypes({@Type(
  value = RASv1Dot1VisaCriterion.class,
  name = "RASv1Dot1VisaCriterion"
)})
public interface OneOfValidatePassportResultMatchedCriterion {
}
```

Note the `property = "type"` in the `@JsonTypeInfo` annotation. After https://github.com/DataBiosphere/terra-external-credentials-manager/pull/227, the same class was generated as:
```
@JsonTypeInfo(
  use = Id.NAME,
  include = As.PROPERTY,
  property = ""
)
@JsonSubTypes({@Type(
  value = RASv1Dot1VisaCriterion.class,
  name = "RASv1Dot1VisaCriterion"
)})
public interface OneOfValidatePassportResultMatchedCriterion {
}
```
The empty string set as the `property` value meant TDR was unable to determine which class the JSON was meant to deserialize into and led to errors like
```
com.fasterxml.jackson.databind.exc.InvalidTypeIdException: Could not resolve subtype of [simple type, class bio.terra.externalcreds.model.OneOfValidatePassportResultMatchedCriterion]: missing type id property '@type' (for POJO property 'matchedCriterion')
```

**How:**
This PR redesigns the inheritance of these model classes and introduces the `discriminator` which is designed to help API consumers determine the type of the object when using polymorphism ([source](https://swagger.io/docs/specification/v3_0/data-models/inheritance-and-polymorphism/#discriminator)). The change in the swagger generator that necessitated this fix was about better supporting the use of discriminators and specifically changed the `property` value in the `@JsonTypeInfo` annotation to stop hardcoding `type` and instead use `discriminator.propertyName` ([source](https://github.com/swagger-api/swagger-codegen-generators/pull/1259/files#diff-498a38db7b07d65092d59ecad0b461af287603cf288eb8e6201293bdf221d38f)). 

I tested this by generating and publishing a client from this branch, pulling that client library into TDR on my BEE, and verifying that TDR could successfully gate access to snapshots based on the visas in my passport.

[CORE-223]: https://broadworkbench.atlassian.net/browse/CORE-223?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/common/openapi.yml
+++ b/common/openapi.yml
@@ -671,8 +671,7 @@ components:
         criteria:
           type: array
           items:
-            oneOf:
-              - $ref: '#/components/schemas/RASv1Dot1VisaCriterion'
+            $ref: '#/components/schemas/VisaCriterion'
 
     ValidatePassportResult:
       type: object
@@ -681,8 +680,7 @@ components:
         valid:
           type: boolean
         matchedCriterion:
-          oneOf:
-            - $ref: '#/components/schemas/RASv1Dot1VisaCriterion'
+          $ref: '#/components/schemas/VisaCriterion'
         auditInfo:
           type: object
           additionalProperties:
@@ -696,6 +694,12 @@ components:
           type: string
         issuer:
           type: string
+      discriminator:
+        propertyName: type
+        mapping:
+          RASv1Dot1VisaCriterion: '#/components/schemas/RASv1Dot1VisaCriterion'
+      oneOf:
+        - $ref: '#/components/schemas/RASv1Dot1VisaCriterion'
 
     RASv1Dot1VisaCriterion:
       allOf:
--- a/service/src/main/java/bio/terra/externalcreds/controllers/OpenApiConverters.java
+++ b/service/src/main/java/bio/terra/externalcreds/controllers/OpenApiConverters.java
@@ -4,10 +4,9 @@ import bio.terra.common.exception.BadRequestException;
 import bio.terra.externalcreds.ExternalCredsException;
 import bio.terra.externalcreds.generated.model.AdminLinkInfo;
 import bio.terra.externalcreds.generated.model.LinkInfo;
-import bio.terra.externalcreds.generated.model.OneOfValidatePassportRequestCriteriaItems;
-import bio.terra.externalcreds.generated.model.OneOfValidatePassportResultMatchedCriterion;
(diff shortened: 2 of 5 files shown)


calypsomatic (approved):
Did not test myself, but your explanation makes sense.

davidangb (approved):
nice sleuthing!

pshapiro4broad (approved):
This looks fine but it would be nice to see a unit test which verifies this fix. Openapi's java support for `oneOf` and `discriminator` has been shaky in the past and I wouldn't be surprised if it broke again in the future.

marctalbott:
> This looks fine but it would be nice to see a unit test which verifies this fix. Openapi's java support for `oneOf` and `discriminator` has been shaky in the past and I wouldn't be surprised if it broke again in the future.

Agreed. I've filed a separate [ticket](https://broadworkbench.atlassian.net/browse/CORE-375) to add some unit tests around this since I'd like to get this fix out so I can continue the work to validate our RAS test plan and unblock that work.

PR #256, opened 2025-03-07
CORE-332: merge TokenProviderService and PassportProviderService

davidangb:

Code consolidation, in response to https://github.com/DataBiosphere/terra-external-credentials-manager/pull/253#discussion_r1984052010. Specifically:
* all logic previously in `TokenProviderService` moves to `PassportProviderService`; there was already a lot of duplication between these two
* all callers of `TokenProviderService` now call `PassportProviderService`. The largest change is in `OauthApiController.createLink`, where branching logic could be collapsed
* unit tests are similarly consolidated

As of this PR, "PassportProviderService" is now badly named. However, I stopped short of renaming it, as a rename would cascade many lines of change. I am open to renaming in this PR or in a follow-on, if it would make this one easier to review. _Reviewers: should I do the rename in this PR?_


service/src/main/java/bio/terra/externalcreds/services/PassportProviderService.java
```diff
@@ -85,6 +92,21 @@ public LinkedAccountWithPassportAndVisas createLink(
...
+              linkedAccount.getLeft(), linkedAccount.getRight()));
+    } else {
+      linkedAccountService.upsertLinkedAccount(linkedAccount.getLeft());
+      // returns a LinkedAccountWithPassportAndVisas with empty passports and visas
+      return new LinkedAccountWithPassportAndVisas.Builder()
+          .linkedAccount(linkedAccount.getLeft())
+          .build();
+    }
+  }
+
```
davidangb, line 109: this is a new helper method for this PR

service/src/main/java/bio/terra/externalcreds/services/PassportProviderService.java
```diff
@@ -168,6 +191,32 @@ public int refreshExpiringPassports() {
...
+              "The linked account found for user ID: %s and provider: %s has expired. "
+                  + "Please go to the Terra Profile page External Identities tab "
+                  + "to re-link your account for this provider.",
+              userId, provider));
+    }
+    var providerProperties = externalCredsConfig.getProviderProperties(provider);
+    return accessTokenCacheService.getLinkedAccountAccessToken(
+        linkedAccount, new HashSet<>(providerProperties.getScopes()), auditLogEventBuilder);
+  }
+
```
davidangb, line 219: `getProviderAccessToken()` moved over from `TokenProviderService`

service/src/main/java/bio/terra/externalcreds/util/ProviderUtils.java
```diff
@@ -1,15 +1,14 @@
...
 public class ProviderUtils {
-  private static Set<Provider> fenceProviders =
-      Set.of(Provider.FENCE, Provider.DCF_FENCE, Provider.KIDS_FIRST, Provider.ANVIL);
 
   private ProviderUtils() {}
 
-  public static boolean isFenceProvider(Provider provider) {
-    return fenceProviders.contains(provider);
+  public static boolean isPassportProvider(ProviderProperties providerProperties) {
+    return providerProperties.getScopes().contains(GA4GH_PASSPORT_V1_CLAIM);
```
davidangb, line 12: As Doug mentioned in https://github.com/DataBiosphere/terra-external-credentials-manager/pull/253#discussion_r1984052010, we might want a different way to identify which providers create passports, such as a new flag in config. This seems ok for now and hopefully this is encapsulated enough to make it easy to change later.

dvoet (approved):
A+

dvoet:
Instead of a rename, I think the next step is to pull everything up from `PassportProviderService` to `ProviderService`. That can be another PR or in this one or just later, up to you.

kevinmarete (approved):

PR #262, opened 2025-03-12
CORE-332: provider service consolidation

davidangb:

The confusing part of this PR when looking at diffs:
* rename `ProviderService` to `ProviderServiceSupport`
* rename `PassportProviderService` to `ProviderService`
* rename tests correspondingly

Using their new names, the rest of the PR:
* all callers now use `ProviderService` instead of `ProviderServiceSupport`. Since `ProviderService` extends `ProviderServiceSupport`, this imposes no change on callers.
* `ProviderServiceSupport` becomes abstract with a protected constructor so nothing can instantiate it directly

## Review Notes

1. given [this comment](https://github.com/DataBiosphere/terra-external-credentials-manager/pull/256#issuecomment-2707406877), would you rather see `ProviderService` and `ProviderServiceSupport` consolidated into a single (large-ish) class, or do you like this approach with using a "support" abstract class?
2. Sonar is reporting a bunch of issues on code I haven't touched other than renaming classes; I am choosing to ignore those. If those are worth fixing I'd prefer to do it in a follow-on PR.


service/src/main/java/bio/terra/externalcreds/services/ProviderServiceSupport.java
davidangb, line 1: this is not new; it was renamed from `ProviderService`. The only code change is making it abstract.

service/src/test/java/bio/terra/externalcreds/services/PassportProviderServiceTest.java
davidangb, line 1: renamed to ProviderServiceTest with no code changes

service/src/test/java/bio/terra/externalcreds/services/ProviderServiceSupportTest.java
davidangb, line 1: renamed from PassportProviderServiceTest with no code changes

service/src/test/java/bio/terra/externalcreds/services/ProviderServiceTest.java
davidangb, line 1: renamed from PassportProviderServiceTest with no code changes

service/src/test/java/bio/terra/externalcreds/services/ProviderServiceSupportTest.java
davidangb, line 1: renamed from ProviderServiceTest with no changes

dvoet (commented):
Given there is only one subclass of ProviderServiceSupport, what is the benefit of keeping separate from ProviderService? The separation seems arbitrary.

davidangb:
closing in favor of #263

PR #264, opened 2025-03-12
CORE-380 use library to call revoke refresh tokens

dvoet:
Jira: https://broadworkbench.atlassian.net/browse/CORE-380

This works for Gen3 (which are the only live providers). RAS seems to have other problems (ticket open with the RAS team). Github seems to do something non standard. This PR advances things so that the actual production providers work.

What:

  \<For your reviewers' sake, please describe in a sentence or two what this PR is accomplishing (usually from the users' perspective, but not necessarily).\>

Why:

  \<For your reviewers' sake, please describe in ~1 paragraph what the value of this PR is to our users or to ourselves.\>

How:

  \<For your reviewers' sake, please describe in ~1 paragraph how this PR accomplishes its goal.\>

  \<If the PR is big, please indicate where a reviewer should start reading it (i.e. which file or function).\>


service/src/main/java/bio/terra/externalcreds/services/OAuth2Service.java
```diff
@@ -136,4 +147,36 @@ public OAuth2User getUserInfo(ClientRegistration providerClient, OAuth2AccessTok
...
+      var response = req.toHTTPRequest().send();
+      log.info(
+          "Token revocation request for user [{}], provider [{}] returned status [{}] with the result: [{}]",
+          linkedAccount.getUserId(),
+          linkedAccount.getProvider().toString(),
+          response.getStatusCode(),
+          response.getContent());
+    } catch (IOException e) {
+      log.error("revoke refresh token failure", e);
+    }
```
davidangb, line 180: is there anything sensitive in the successful response content or stack trace we don't want to log?

davidangb (approved):
one question inline, proactive :+1:. Didn't test this myself.

calypsomatic (approved):

service/src/main/java/bio/terra/externalcreds/services/OAuth2Service.java
dvoet: good thought, but I don't think so

PR #290, opened 2025-07-15
[CORE-560] Get linked era user id when getting RAS provider info

samanehsan:
Jira: https://broadworkbench.atlassian.net/browse/CORE-560

What: Update the GET provider details endpoint to include the ERA commons user id for the RAS provider (if their account is linked)

Why: Make it easier for a user to know if their ERA commons account is linked to the account they used to link with RAS

How: When getting RAS linked account info, get user info from RAS and extract the era userid from the "federated_identities" attribute

Testing:
I tested this locally with my dev RAS account to confirm that I can at least get `userInfo` when I call the `getLink` endpoint. It is currently not possible to link an ERA Commons account in dev or staging, so it does not have the "federated_identities" attribute. Instead I wrote unit tests based on the expected format.

### DIFF
--- a/service/src/main/java/bio/terra/externalcreds/controllers/OauthApiController.java
+++ b/service/src/main/java/bio/terra/externalcreds/controllers/OauthApiController.java
@@ -43,7 +43,20 @@ public record OauthApiController(
   public ResponseEntity<LinkInfo> getLink(Provider provider) {
     var samUser = samUserFactory.from(request);
     var linkedAccount = linkedAccountService.getLinkedAccount(samUser.getSubjectId(), provider);
-    return ResponseEntity.of(linkedAccount.map(OpenApiConverters.Output::convert));
+    Optional<LinkInfo> linkInfo = linkedAccount.map(OpenApiConverters.Output::convert);
+    if (linkInfo.isPresent() && provider == Provider.RAS) {
+      var auditLogEventBuilder =
+          new AuditLogEvent.Builder()
+              .provider(provider)
+              .userId(samUser.getSubjectId())
+              .clientIP(request.getRemoteAddr());
+      var era_user_id =
+          providerService.getLinkedEraIdentity(provider, linkedAccount.get(), auditLogEventBuilder);
+      if (era_user_id != null) {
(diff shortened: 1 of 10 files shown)


service/src/main/java/bio/terra/externalcreds/controllers/OauthApiController.java
```diff
@@ -43,7 +43,20 @@ public ResponseEntity<List<String>> listProviders() {
...
-    return ResponseEntity.of(linkedAccount.map(OpenApiConverters.Output::convert));
+    Optional<LinkInfo> linkInfo = linkedAccount.map(OpenApiConverters.Output::convert);
+    if (linkInfo.isPresent() && provider == Provider.RAS) {
+      var auditLogEventBuilder =
+          new AuditLogEvent.Builder()
+              .provider(provider)
+              .userId(samUser.getSubjectId())
+              .clientIP(request.getRemoteAddr());
+      var era_user_id =
+          providerService.getLinkedEraIdentity(provider, linkedAccount.get(), auditLogEventBuilder);
```
calypsomatic, line 54: I don't know this code very well, but is there a chance of this throwing some kind of error, and if so, would we want to handle it or just let it throw?

samanehsan: Oh you're right! This could throw an error if getting the user info from RAS fails...I think it's the connection refused error but I wonder if I should just add a more general try/catch around this line and then log the exception?

calypsomatic: Ah I see it's defined down below... same question, passed down to that method then maybe

davidangb (approved):
I agree with Bria's comment about error trapping/resilience; otherwise :+1: and thank you!

service/src/main/java/bio/terra/externalcreds/controllers/OauthApiController.java
samanehsan: @calypsomatic, I updated the error handling if you want to take another look 👀

calypsomatic (approved):

PR #292, opened 2025-07-18
CORE-612: Stop pact tests

davidangb:
Jira: CORE-612

Deletes all pact tests and the GHA that ran them.

The single existing pact involving ECM has CBAS as a consumer; we don't need to keep testing that.
Screenshot 18-07-2025 at 14 51


client-resttemplate/src/main/java/bio/terra/externalcreds/pact/ProviderStates.java
```diff
@@ -15,6 +15,7 @@
 
 package bio.terra.externalcreds.pact;
 
+@Deprecated(forRemoval = true, since = "2025-07-18")
```
davidangb, line 18: It is probably safe to just delete this class entirely; but since it's packaged as part of the ECM client I am being conservative about breakage and am deprecating it instead of deleting.

samanehsan (approved):

aednichols (approved):

PR #304, opened 2025-12-10
[CTM-284] Update RAS dev and staging scopes

samanehsan:
Jira: https://broadworkbench.atlassian.net/browse/CTM-284

What:
Update the existing `federated_identities` scope to `federated_identities_ial2` for the RAS dev and staging clients. Previously all RAS clients had the same scopes in every environment, so in order to make this change the scopes are now defined per-environment.

Why:
We have a new RAS client that supports the NIST Identity Assurance Level 2 (IAL2) standard but it requires a different scope. 

How:
Update the ECM providers config, just for dev and staging RAS.

### DIFF
--- a/service/src/main/resources/providers.yml
+++ b/service/src/main/resources/providers.yml
@@ -121,7 +121,6 @@ externalcreds:
       clientId: "${RAS_CLIENT_ID}"
       clientSecret: "${RAS_CLIENT_SECRET}"
       externalIdClaim: "email"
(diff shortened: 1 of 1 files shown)


mbaumann-broad (approved):
LGTM @samanehsan, thank you for your work on this!
I defer to @TomConner for any additional comments he may have.

marctalbott (approved):

jgainerdewar (approved):

PR #317, opened 2026-04-29
[CTM-443] Add RAS support API

calypsomatic:
Jira: https://broadworkbench.atlassian.net/browse/CTM-443

  - Adds a new admin API endpoint GET `/api/admin/v1/ras/user/{userId}` for RAS support troubleshooting
  - Updates the admin authorization check to fall back to Sam's resource type admin permission check
  - Bumps sam-client from v0.0.208 to v0.0.440

  New endpoint: `getRasSupportInfo`

  Returns a snapshot of a user's RAS state useful for troubleshooting linked identity and passport issues. Returns 404 if the user has no RAS link.

  Response structure:

  - link (always present): the user's RAS linked account — external user ID (i.e. RAS email), link expiry timestamp, whether it's authenticated, and whether it's
  expired
  - passport (optional, absent if no passport): passport expiry and whether it's expired
  - dbgapPermissions (optional, empty array if no passport): the decoded dbGaP permissions from the user's passport visas — each entry has phsId, consentGroup, and
  role
Example response:
```
{
  "link": {
    "externalUserId": "test@ras.example.com",
    "linkExpires": "2026-05-15T22:21:01.913+00:00",
    "authenticated": true,
    "expired": false
  }
}
```

  **Admin authorization changes**

  Previously, admin access was gated solely on the caller's email being in ECM's authorizedAdmins config list.

  The check now falls back to Sam's resource type admin permission if the email check fails: it calls Sam's `resourceTypeAdminPermission("user", "admin_read_summary_information")` using the caller's bearer token. Sam has an empty user resource type that supports resource-type-level admin permissions without  requiring any specific resource instance. If the Sam call returns false or throws for any reason, it falls through to a 403.

  The email allowlist remains the primary check and is unaffected — Sam is only called when it fails.


service/src/main/java/bio/terra/externalcreds/controllers/AdminApiController.java
```diff
@@ -79,11 +85,37 @@ public ResponseEntity<List<Object>> getVisas(
...
+    try {
+      if (samAdminDAO.resourceTypeAdminPermission(
+          samUser.getBearerToken().getToken(), "user", "admin_read_summary_information")) {
+        return;
+      }
+    } catch (Exception e) {
+      log.warn("Sam resourceTypeAdminPermission check failed", e);
     }
+    throw new ForbiddenException("Admin permissions required");
   }
```
marctalbott: It might be worth adding separate authz checks for read vs write admin operations. Most of the ECM admin endpoints are read-only anyways, but not sure if admin unlinking should be gated by the `admin_read_summary_information` action too. 

It's a trusted group of internal users that gain this access so it's not a huge gap or anything

marctalbott (approved):

aednichols (approved):

marctalbott (approved):

service/src/main/java/bio/terra/externalcreds/controllers/AdminApiController.java
calypsomatic: Done

